PR #44, opened 2015-03-18
Add additional JAVA-Doc param

KaiVolland:
JAVA 8 uses a strict doclint. To avoid errors when building we added an additional param configuration.


buehner:
Thank you!

PR #46, opened 2015-06-25
Update dependencies

buehner:
I made some cleanup and updated the versions of dependencies.

Please review and merge if travis is happy.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ jdk:
 
 before_script:
   - cd src/
-  - mvn --batch-mode --threads 2.0C -DskipTests=true -Dmaven.javadoc.skip=true install
+  - mvn --batch-mode --threads 2.0C -DskipTests=true install
 
 script:
   - mvn --batch-mode --threads 2.0C --show-version clean test
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -79,7 +79,6 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
-        <jdk.version>1.8</jdk.version>
         <tomcat.version>8.0.23</tomcat.version>
 
         <!-- Core -->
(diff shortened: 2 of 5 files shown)


marcjansen:
This looks great, @buehner.

Since travis is hanging at the moment, you may merge at your own will. I am confident this doesn't break anything.

chrismayer:
Thanks for this @buehner.
Is it an option to declare the versions with placeholders, so we always get the latest bugfix for example? OR do you prefer setting the version manually? Maybe you will have a look at this (if not known already) http://stackoverflow.com/a/1172371

marcjansen:
Plus one for version ranges.

marcjansen:
The failing builds actually hint something is wrong and are not due to downtime of travis, I guess.

buehner:
Thx for the hint @chrismayer , i will have a look at it.

The build fails on Java 7 as it was not very clever to use 1.8 for the ${jdk.version} property value in the pom.xml.

I will try to add some fixing commits soon.

buehner:
7eb3730 should fix travis by using profiles for different JDK versions.

I also experimented a bit with the version ranges suggested by @chrismayer .
But they did not always behave as expected.

For example i used the range [4.0.0,5.0.0) for hibernate. According to the docs this should always resolve to the latest 4.x.x version, i.e. 4.0.0 <= x < 5.0.0

Unfortunately this resolves to 5.0.0.CR1, which lets the DAO tests fail.

I will have a look at this later. Maybe the the maven versions plugin is a more stable solution: http://www.mojohaus.org/versions-maven-plugin/

As travis is happy now, it would be great if someone could merge!

buehner:
I pushed one more commit (120821e) that will enable the build of javadocs on travis. This is important as there were changes between Java 1.7 and 1.8 (Xdoclint)

marcjansen:
Nice!

PR #48, opened 2015-08-19
Add a profile to make use of UMLgraph for javadocs

buehner:
The new profile can be activated with the CLI argument `-Pumlgraph`, e.g.
`mvn clean package -Pumlgraph`. To make this work, a dependency has to be installed on the system: graphviz. One could use `sudo apt-get install graphviz` to install this.

The UMLGraph software will produce UML class diagram pictures, which will automatically be included in the javadocs. Options for the generation of the diagrams can be found here: http://www.umlgraph.org/doc/cd-opt.html

umlgraph

### DIFF
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -157,6 +157,31 @@
                 <javadoc.opts>-Xdoclint:none</javadoc.opts>
             </properties>
         </profile>
+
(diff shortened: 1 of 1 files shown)


marcjansen:
Gogogo!

PR #54, opened 2015-10-28
Use new packages for layout and module models

buehner:
This PR contains a simple refactoring of the model package structure. As we have a growing number of models, all module classes and all layout classes have been moved to separate (sub-)packages.


KaiVolland:
Nice. :dancers:

PR #55, opened 2015-10-28
Replace deprecated annotation

buehner:
The @TransactionConfiguration annotation is deprecated since spring 4.2.2 and has been replaced with the (recommended) @Rollback annotation.


marcjansen:
:8ball:

PR #56, opened 2015-10-28
Replace deprecated annotation in ACL test

buehner:
The same as in #55 for another test. I missed this in #55


marcjansen:
Yes, please!

PR #57, opened 2015-10-28
Use latest versions of EXT-direct and classgenerator

buehner:
This will add EXT 6 support. Just have a look at https://github.com/terrestris/shogun2/wiki/Basic-concepts#extdirect to understand how you can test this.


dnlkoch:
Great!

PR #62, opened 2015-10-31
Add a ButtonModule class

KaiVolland:
The buttonmodule contains information about apperance and action.

The action of the button is define via enum. It address the buttons contained in
the BasiGX Package (https://github.com/terrestris/BasiGX/tree/master/src/view/button).

Additional there are two more generic types:
* "OPENMODULEWINDOW" (property connectedModule required)
    * Opens any module in an Ext.window.Window.
* "TOGGLEINTERACTION" (property interaction required)
    * Toggles the functionality of an OpenLayers 3 interaction.

### DIFF
--- a/src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/module/Button.java
+++ b/src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/module/Button.java
@@ -4,17 +4,12 @@
 package de.terrestris.shogun2.model.module;
 
 import javax.persistence.Entity;
-import javax.persistence.EnumType;
-import javax.persistence.Enumerated;
 import javax.persistence.Table;
 
 import org.apache.commons.lang3.builder.EqualsBuilder;
 import org.apache.commons.lang3.builder.HashCodeBuilder;
 import org.apache.commons.lang3.builder.ToStringBuilder;
 import org.apache.commons.lang3.builder.ToStringStyle;
-
-import com.fasterxml.jackson.annotation.JsonCreator;
-import com.fasterxml.jackson.annotation.JsonValue;
 
 /**
  * The Image Module is the Ext JS representation of an HTML img element.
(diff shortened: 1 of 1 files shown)


src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/module/Button.java
```diff
@@ -0,0 +1,290 @@
...
+		ADDWMS("addwms"),
+		COORDINATETRANSFORM("coordinatetransform"),
+		HELP("help"),
+		HSI("hsi"),
+		MEASURELINE("measureline"),
+		MEASUREARE("measurearea"),
+		PERMALINK("permalink"),
+		TOGGLELEGEND("togglelegend"),
+		ZOOMIN("zoomin"),
+		ZOOMOUT("zoomut"),
```
buehner: I think this should be ZOOMOUT("zoomout")

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/module/Button.java
```diff
@@ -0,0 +1,290 @@
...
+	 */
+	public static enum ButtonActionType {
+		OPENMODULEWINDOW("openmodulewindow"),
+		TOGGLEINTERACTION("toggleinteraction"),
+		ADDWMS("addwms"),
+		COORDINATETRANSFORM("coordinatetransform"),
+		HELP("help"),
+		HSI("hsi"),
+		MEASURELINE("measureline"),
+		MEASUREARE("measurearea"),
```
buehner: I think this should be MEASUREAREA("measurearea")

buehner:
Thx @KaiVolland 

The button seems nice,but i think the enum ButtonActionType should be refactored to a String ButtonActionType as we currently can not be sure that this will be a fix set of values.

We should also fix the coveralls problem before merging this.

KaiVolland:
Refactored the ButtonActionType.

marcjansen:
Please change the PR title to e.g. "Add a ButtonModule class"

PR #63, opened 2015-10-31
Switch to jacoco as code coverage tool

marcjansen:
… it seems this will make fewer problems wrt Java8.


KaiVolland:
Cool! Seems to be working with JAVA 8 now.

PR #65, opened 2015-11-02
Module properties

dnlkoch:
This PR introduces the property `xtype` to the model Module and may be used to store the the client side class name (or shorthand name in case of ExtJS).

### DIFF
--- a/src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/module/Module.java
+++ b/src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/module/Module.java
@@ -53,6 +53,11 @@ public abstract class Module extends PersistentObject {
 	private String name;
 
 	/**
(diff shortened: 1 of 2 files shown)


KaiVolland:
Some thoughts  to the property name inline. :top: :dancers:

buehner:
@KaiVolland Calling it `type` could be confusing because the [Layout class](https://github.com/dnlkoch/shogun2/blob/f38f450a8e80df249dde44c55c61727493176e40/src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/layout/Layout.java#L52) (which is related to modules) already has a "type" property.

For me, `xtype` is fine. Otherwise i would propose something like `componentType`.

(Using something like `alias` would not be clever as it is a reserved word in SQL, which leads to problems as hibernate is in use.)

What's your opinion, @marcjansen ?

marcjansen:
Xtype is fine.

PR #68, opened 2015-11-05
Add default layer bean

dnlkoch:
Within this PR you will find a new default bean containing the representation of a `de.terrestris.shogun2.model.layer.Layer` object including some fixes in the corresponding models.


marcjansen:
I like it. Nice work.

PR #71, opened 2015-11-06
Integrated mapcontrols into map module, removed interactions

weskamm:
Interactions have been removed from the models as they will get used in submodules like e.g. a button, as discussed with @dnlkoch et al.
MapControls are now integrated in the map module, example added to content-initialize bean.


dnlkoch:
Good work, buddy!

awesome

PR #73, opened 2015-11-07
Avoid LazyInitializationException with open session in view

buehner:
We were using the `FetchType.EAGER` configuration for (almost) all of our associations to ensure that we will not get a `LazyInitializationException` in the web layer, i.e. when Jackson serializes the Java objects to JSON.

Since version 4.2, Spring provides a `OpenSessionInViewFilter` that makes hibernate sessions available in the web/view layer, so that associated objects can be loaded lazily there.

http://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/orm/hibernate5/support/OpenSessionInViewFilter.html

There is a controversial discussion whether this pattern is good practice or not:

* https://stackoverflow.com/questions/1103363/why-is-hibernate-open-session-in-view-considered-a-bad-practice
* http://blog.jhades.org/open-session-in-view-pattern-pros-and-cons/
* https://blog.frankel.ch/the-opensessioninview-antipattern

In my eyes, some of the criticisms do not apply to us. For example, we are not running SHOGun2 on *distributed* JVMs. As far as i understand, one of the main differences between loading `EAGER` and using an `OpenSessionInViewFilter` is the way the database is being queried: While associations with `EAGER` loading will be fetched with SQL joins, the `OpenSessionInViewFilter` approach will lead to `N + 1` select queries. This could become a performance issue, but in such a scenario one could still use *clever* `EAGER` annotations for certain associations to enhance the performance.

I would like to give the `OpenSessionInViewFilter` approach a chance, because it's simple, saves code, will preserve us from unexpected `LazyInitializationExceptions` and the performance of future business logic implementations in this project could benefit from the `LAZY` loading approach in the (transactional) service layer.

What's your opinion?

### DIFF
--- a/src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/Application.java
+++ b/src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/Application.java
@@ -4,7 +4,6 @@ import java.util.Locale;
 
 import javax.persistence.Column;
 import javax.persistence.Entity;
-import javax.persistence.FetchType;
 import javax.persistence.Inheritance;
 import javax.persistence.InheritanceType;
 import javax.persistence.ManyToOne;
@@ -79,7 +78,7 @@ public class Application extends PersistentObject {
 	/**
 	 *
 	 */
-	@ManyToOne(fetch = FetchType.EAGER)
+	@ManyToOne
 	@Cascade(CascadeType.SAVE_UPDATE)
 	private Viewport viewport;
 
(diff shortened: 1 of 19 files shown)


marcjansen:
I have zero experience with this, but am not opposed against your solution. I am plus 1 on merging.

KaiVolland:
Sounds good to me. As you read a lot about it i think you can decide if its usefull or not. Merge it if you like. :dancers:

dnlkoch:
PR was moved to https://github.com/terrestris/shogun2/pull/80.

PR #74, opened 2015-11-08
Convert values of properties map

buehner:
Until now, the `properties` map of the `Module` class was of type `Map<String, String>`. The JSON representation of such maps holds only string values - even for numbers or booleans:

```json
"properties": {
    "integervalue": "42",
    "doublevalue": "-17.42",
    "booleanvalue": "true",
    "stringvalue": "foo"
}
```

This PR changes the value type of the `properties` map to `Map<String, Object>` to allow the administration of arbitrary types (`Object`). As hibernate cannot handle such maps just like that, a JPA converter has been implemented (`PropertyValueConverter`). This converter will store all values as their string representation in the database. When such a string value is read from the database, the converter tries to parse it as the best matching Java Type, e.g.

`"foo" -> "foo"`
`"42" -> 42`
`"17.42" -> 17.42`
`"true" -> true`

This leads to the following JSON representation:

```json
"properties": {
    "integervalue": 42,
    "doublevalue": -17.42,
    "booleanvalue": true,
    "stringvalue": "foo"
}
```

Of course, this approach has some disadvantages:

* There will never be string values with digits only in our JSON (as they would be parsed as numbers)
* The strings "true" and "false" would always be converted to boolean values.

I would like to merge this PR as I don't think that this is a major problem for now. A client could still convert the primitives to strings, if necessary. If really needed, we could also think about a concept to allow "true" or "12345" strings  by introducing some kind of escape char, prefix or suffix.

This PR also contains tests for the `PropertyValueConverter`

### DIFF
--- a/src/shogun2-core/shogun2-model/pom.xml
+++ b/src/shogun2-core/shogun2-model/pom.xml
@@ -17,13 +17,9 @@
 
     <dependencies>
         <dependency>
-            <groupId>org.hibernate</groupId>
-            <artifactId>hibernate-core</artifactId>
-        </dependency>
-
-        <dependency>
-            <groupId>org.apache.commons</groupId>
-            <artifactId>commons-lang3</artifactId>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>shogun2-util</artifactId>
+            <version>${project.version}</version>
         </dependency>
 
         <dependency>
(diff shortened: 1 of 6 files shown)


marcjansen:
I think this is OK. The performance penalty when serializing / deserializing can probably be ignored, and I like that the better semantics in the returned JSON. Plase merge, @buehner

buehner:
Thx

PR #76, opened 2015-11-16
Change class type for models Module and CompositeModule

dnlkoch:
This PR suggests to handle the models `Module` and `CompositeModule` as classes instead of abstract classes. Having this set, we will be able to use these as models for simple modules like a plain ExtJS panel.


KaiVolland:
This could be very helpful. Plz merge. :top: :+1: :dancers:

dnlkoch:
Thx! Will merge.

PR #77, opened 2015-11-16
Updated models LayerAppearance and WmsTileGrid

dnlkoch:
Changes in module `LayerAppearance`: 
* The property `type` is removed because the layer type is already set in the `Layer` model.
* The properties `maxScale` and `minScale` are replaced with `maxResolution` and `minResolution` as the ol.layer hasn't support for scales in this context and we are now able to re-use any given `Resolution` entity here.

Changes in module `WmsTileGrid`:
* Property `tileGridResolutions` is added as it is required in the ol.tilegrid class.
* Add property `key` to explicitly set the ol.tilegrid type.


KaiVolland:
Good Work Big-D! :dancers: merge it

dnlkoch:
Thx! Will merge.

PR #81, opened 2015-11-18
Model cleanup

buehner:
As the `Module` and the `CompositeModule` classes are not abstract anymore since #76, i removed all their subclasses that did not provide any additional attributes (i.e. where the above mentioned non-abstract classes can be used now).

I also added missing `@Table` annoations to these classes.

I could successfully run the adapted webapp-archetype and start the application afterwards!

Please review and merge. Thx


dnlkoch:
te

PR #83, opened 2015-11-19
Add tests

buehner:
This PR is ready to be merged now.

I added missing tests that will increase the coverage to more than 70%.

A bug that was found during the testing has also been fixed in this PR:

* The `findByCriteriaWithSortingAndPaging` method of `GenericHibernateDao` did not return a correct total count in all cases as the criteria were not respected in when counting. See 38dae57b


dnlkoch:
+1 

Good work!

PR #85, opened 2015-12-02
Encrypt passwords

buehner:
This PR enables password encryption. Passwords will be encrypted and checked by using the `BCryptPasswordEncoder` of the spring framework.

* Encryption is done when a new user is being created
* The `Shogun2AuthenticationProvider` will check if a given raw password string (e.g. from a login mask) matches the password hash from the database.

Please review and merge.


buehner:
After 30ae54c (fix build) the coverage/coveralls failed because the coverage decreased (-2.3%).
I configured coveralls to fail only, if:

* coverage of project gets below 50 %
* coverage decreases more than 5 %

Afterwards i restared the travis build manually.

weskamm:
We need to remember, as soon as we have a web-interface for updating existing users, this logic will also be needed there. I am not sure but would assume that a simple `saveorupdate` would not handle the encryption at all.
Otherwise, this looks good to me!

buehner:
:+1:

PR #87, opened 2015-12-04
Handle role hierarchy

buehner:
Let the `Shogun2AuthenticationProvider` explicitly add all **reachable** roles of the role hierarchy (as granted authorities) to the auth result of a principal/user. This is necessary if we want to make use of one of these roles when configuring the `AclAuthorizationStrategyImpl`.

On top of that, some minor changes (like log level or comments) have been made.

Please review and merge. Thx


src/shogun2-init/src/main/java/de/terrestris/shogun2/init/ContentInitializer.java
```diff
@@ -150,7 +146,7 @@
 	 * Using the {@link Resource} annotation as
 	 * recommended on http://stackoverflow.com/a/22463219
 	 */
-	@Resource(name = "defaultRoles")
```
dnlkoch: Comment isn't in sync with the code anymore (same below)

buehner:
Thx for the comment regarding the outdated comments. I pushed the fix and wil merge afterwards. Thx

PR #88, opened 2015-12-04
Default security namespace

buehner:
In webapp archetype:

Change the default namespace in security config. Also reordered/formatted the security config.


dnlkoch:
Looks good to me, I'll merge!

PR #91, opened 2015-12-08
Initial commit of the REST Controllers

KaiVolland:
Including
- AbstractRestController
- ApplicationRestController
- UserRestController

### DIFF
--- a/src/shogun2-web/src/main/java/de/terrestris/shogun2/rest/AbstractRestController.java
+++ b/src/shogun2-web/src/main/java/de/terrestris/shogun2/rest/AbstractRestController.java
@@ -4,6 +4,8 @@ import java.util.List;
 
 import org.apache.log4j.Logger;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RequestMapping;
@@ -33,8 +35,8 @@ public abstract class AbstractRestController<E extends PersistentObject> {
 	 * @return
 	 */
 	@RequestMapping(method = RequestMethod.GET)
-	public List<E> findAll() {
-		return this.service.findAll();
+	public ResponseEntity<List<E>> findAll() {
+		return new ResponseEntity<List<E>>(this.service.findAll(), HttpStatus.OK);
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


buehner:
Thx @KaiVolland 

Here are some points that i would like to see changed before merging:

* The required changes for the servlet mappings in `web.xml` (webapp archetype) are missing.
* Remove all `System.out.println(..)` and use a logger instance instead
* Replace variables named `json` to something like `entity` (as REST is not always JSON and even in our specific case the JSON has already been deserialized to java objects at this point)
* Regarding the `update(..)` method:
  * Use `PUT` instead of `POST`
  * Handle HTTP Status Codes in case of exceptions or empty results (see links below)
  * The `saveOrUpdate` method from the `AbstractCrudService` always expects a *complete* entity object, i.e. this method is not capable to handle updates of partial resources only (see second link below). For this purpose we should extend the DAO/service Layer to support something like *merge*. For now, it should be OK for us to expect a *complete* resource payload, but then we should **not** make use of something like `BeanUtils.copyProperties(..)`. I suggest the following: Check if the `id` of the passed entity equals the value of the id-`PathVariable`. If true: simply call `this.service.saveOrUpdate(entity);
`. It does not make sense to fetch the whole object from database first if we already have the *complete* object to update.

Regarding HTTP Status Codes and HTTP methods, this is possibly a good orientation: http://www.restapitutorial.com/lessons/httpmethods.html

This article also mentions that the `HTTP PATCH` method can be used to update only partial resources: http://restful-api-design.readthedocs.org/en/latest/methods.html

All in all this seems to be a very nice and generic approach for a RESTful web interface. Thank you :+1:

KaiVolland:
- [x] Adjust webapp-archetype in web.xml and servlet.xml
- [ ] Remove all System.out.println(..) and use a logger instance instead
- [ ] Replace variables named json to something like entity (as REST is not always JSON and even in our specific case the JSON has already been deserialized to java objects at this point)
- [ ] Adjust update method as @buehner suggestes

buehner:
Please use the `debug` or `trace` log level. I think `LOG.info(..)` is a bit to much for this class ;-)

KaiVolland:
I think the most is done.

Currently i'm not using the status codes from http://www.restapitutorial.com/lessons/httpmethods.html.
I'll implement them (or similar codes) tomorrow.

KaiVolland:
Added a basic test.
Some things should be changed/improved:
- It tests the AbstractController with an Application instead of using the real abstract implementation <E>. There should be a way to do this better.
- Only the GET-Request to findAll applications is test. We need to add the rest.

Maybe someone more JAVA-experienced should push this further.

 :santa:

buehner:
Thx for the adaptions and the test. I will merge now.

PR #96, opened 2015-12-14
REST exception and HTTP status code handling

buehner:
I adapted the following aspects of the current REST implementation:

* Logging
* Use plural (e.g. `applications` instead of `application`) for the REST web interface as this is the most common way
* Return more meaningful HTTP status codes (see below)

I added a `loadById` method to the DAO and service, which is nice to have, e.g. to load (a proxy) of an object when deleting by id. This is more performant than using `findById`, which immediately accesses the database.

| Path  | HTTP method  | HTTP Status Code(s)  |
|---|---|---|
| `rest/{entity}`  | GET  | `200 OK` (with filled or empty array) |
| `rest/{entity}/{id}`  | GET  | `200 OK`<br>`404 NOT FOUND`, if entity not found or access is denied  |
|  `rest/{entity}` | POST  | `200 OK`<br>`400 BAD REQUEST` in case of exception |
|  `rest/{entity}/{id}` | PUT  | `200 OK`<br>`404 NOT FOUND` in case of exception (e.g. if object with `{id}` does not exist)<br>`400 BAD REQUEST`, if `{id}` does not match the ID in the payload |
|  `rest/{entity}/{id}` | DELETE  | `204 NO CONTENT` (i.e. empty response) in case of successful deletion<br>`404 NOT FOUND` in case of exception (e.g. if object with `{id}` does not exist) |

Maybe i will push some more tests if you wait with merging ;-)

### DIFF
--- a/src/shogun2-core/shogun2-dao/src/main/java/de/terrestris/shogun2/dao/GenericHibernateDao.java
+++ b/src/shogun2-core/shogun2-dao/src/main/java/de/terrestris/shogun2/dao/GenericHibernateDao.java
@@ -38,12 +38,42 @@ public abstract class GenericHibernateDao<E extends PersistentObject, ID extends
 	@Autowired
 	private SessionFactory sessionFactory;
 
-	public Session getSession() {
+	/**
+	 * Obtains the current session.
+	 *
+	 * @return
+	 */
+	private Session getSession() {
 		return sessionFactory.getCurrentSession();
 	}
 
+	/**
+	 * Return the real object from the database. Returns null if the object does
+	 * not exist.
+	 *
+	 * @param id
+	 *
+	 * @see http://www.mkyong.com/hibernate/different-between-session-get-and-session-load/
+	 *
+	 * @return The object from the database or null if it does not exist
+	 */
 	public E findById(ID id) {
 		return (E) getSession().get(clazz, id);
+	}
+
+	/**
+	 * Return a proxy of the object (without hitting the database). This should
+	 * only be used if it is assumed that the object really exists and where
+	 * non-existence would be an actual error.
+	 *
+	 * @param id
+	 *
+	 * @see http://www.mkyong.com/hibernate/different-between-session-get-and-session-load/
+	 *
+	 * @return
+	 */
+	public E loadById(ID id) {
+		return (E) getSession().load(clazz, id);
 	}
 
 	/**
--- a/src/shogun2-core/shogun2-service/src/main/java/de/terrestris/shogun2/service/AbstractCrudService.java
+++ b/src/shogun2-core/shogun2-service/src/main/java/de/terrestris/shogun2/service/AbstractCrudService.java
@@ -10,16 +10,16 @@ import de.terrestris.shogun2.model.PersistentObject;
 
 /**
  * This abstract service class provides basic CRUD functionality.
- * 
+ *
  * @author Nils Bühner
  * @see AbstractDaoService
- * 
+ *
  */
 public abstract class AbstractCrudService<E extends PersistentObject> extends
 		AbstractDaoService<E> {
 
 	/**
-	 * 
+	 *
 	 * @param e
 	 * @return
 	 */
@@ -30,6 +30,8 @@ public abstract class AbstractCrudService<E extends PersistentObject> extends
 	}
 
 	/**
+	 * Return the real object from the database. Returns null if the object does
+	 * not exist.
 	 * 
 	 * @param id
 	 * @return
@@ -40,7 +42,20 @@ public abstract class AbstractCrudService<E extends PersistentObject> extends
 	}
 
 	/**
+	 * Return a proxy of the object (without hitting the database). This should
+	 * only be used if it is assumed that the object really exists and where
+	 * non-existence would be an actual error.
 	 * 
+	 * @param id
+	 * @return
+	 */
+	@PostAuthorize("hasPermission(returnObject, 'READ')")
+	public E loadById(int id) {
+		return dao.loadById(id);
+	}
+
+	/**
+	 *
 	 * @return
 	 */
 	@PostFilter("hasPermission(filterObject, 'READ')")
@@ -49,7 +64,7 @@ public abstract class AbstractCrudService<E extends PersistentObject> extends
 	}
 
 	/**
-	 * 
+	 *
 	 * @param e
 	 */
 	@PreAuthorize("hasPermission(#e, 'DELETE')")
--- a/src/shogun2-web/src/main/java/de/terrestris/shogun2/rest/AbstractRestController.java
+++ b/src/shogun2-web/src/main/java/de/terrestris/shogun2/rest/AbstractRestController.java
@@ -40,8 +40,15 @@ public abstract class AbstractRestController<E extends PersistentObject> {
 	 */
 	@RequestMapping(method = RequestMethod.GET)
 	public ResponseEntity<List<E>> findAll() {
-		return new ResponseEntity<List<E>>(this.service.findAll(),
-				HttpStatus.OK);
+		final List<E> resultList = this.service.findAll();
+
+		if (resultList != null && !resultList.isEmpty()) {
+			LOG.trace("Found a total of " + resultList.size()
+					+ " entities of type "
+					+ resultList.get(0).getClass().getSimpleName());
+		}
+
+		return new ResponseEntity<List<E>>(resultList, HttpStatus.OK);
 	}
 
 	/**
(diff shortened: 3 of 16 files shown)


marcjansen:
What do we return on POST or PUT, when e.g. required fields are not present? E.g. fields that cannot be null are missing in the payload.

src/shogun2-core/shogun2-service/src/main/java/de/terrestris/shogun2/service/AbstractCrudService.java
```diff
@@ -40,7 +42,20 @@ public E findById(Integer id) {
...
 
 	/**
+	 * Return a proxy of the object (without hitting the database). This should
+	 * only be used if it is assumed that the object really exists and where
+	 * non-existence would be an actual error.
 	 * 
+	 * @param id
+	 * @return
+	 */
+	@PostAuthorize("hasPermission(returnObject, 'READ')")
```
marcjansen: Why is it PostAuthorize here and PostFilter elsewhere?

src/shogun2-core/shogun2-dao/src/main/java/de/terrestris/shogun2/dao/GenericHibernateDao.java
```diff
@@ -38,15 +38,45 @@ protected GenericHibernateDao(Class<E> clazz) {
...
+	 * non-existence would be an actual error.
+	 *
+	 * @param id
+	 *
+	 * @see http://www.mkyong.com/hibernate/different-between-session-get-and-session-load/
+	 *
+	 * @return
+	 */
+	public E loadById(ID id) {
+		return (E) getSession().load(clazz, id);
```
marcjansen: So will this method throw in case that the id isn't pointing to a valid existing object?

Or will subsequent actions fail loudly?

src/shogun2-web/src/main/java/de/terrestris/shogun2/rest/AbstractRestController.java
```diff
@@ -32,14 +34,21 @@
...
 	 *
 	 * @return
 	 */
 	@RequestMapping(method = RequestMethod.GET)
 	public ResponseEntity<List<E>> findAll() {
-		return new ResponseEntity<List<E>>(this.service.findAll(),
-				HttpStatus.OK);
+		final List<E> resultList = this.service.findAll();
+
+		if (resultList != null && !resultList.isEmpty()) {
```
marcjansen: Super minor: can we somehow check if TRACE log level is currently set? So we skip the whole if-block?

src/shogun2-web/src/main/java/de/terrestris/shogun2/rest/AbstractRestController.java
```diff
@@ -49,25 +58,48 @@
...
+			LOG.error(errorMessagePrefix + "ID value is set to " + id
+					+ ", but MUST be null");
+			return new ResponseEntity<E>(HttpStatus.BAD_REQUEST);
+		}
+
 		try {
-			E created = this.service.saveOrUpdate(entity);
-			LOG.debug("Create " + entity.getClass() + " with body " + entity);
-			return new ResponseEntity<E>(created, HttpStatus.CREATED);
+			entity = this.service.saveOrUpdate(entity);
```
marcjansen: Don't we now know that this will not be a CREATE? Is there a dedicazed `save` in the service?

src/shogun2-core/shogun2-service/src/main/java/de/terrestris/shogun2/service/AbstractCrudService.java
buehner: Filter is used to filter collections: https://docs.spring.io/spring-security/site/docs/current/reference/html/el-access.html#el-pre-post-annotations

src/shogun2-web/src/main/java/de/terrestris/shogun2/rest/AbstractRestController.java
```diff
@@ -49,25 +58,48 @@
...
+			LOG.trace("Created " + simpleClassName + " with ID " + entity.getId());
```
marcjansen: Wrong log message, right?

src/shogun2-core/shogun2-dao/src/main/java/de/terrestris/shogun2/dao/GenericHibernateDao.java
buehner: This will NOT throw immediately. So you will later get an exception, if you try to access it, but it does not exist. In our case, we catch this exception in the delete method of `AbstractRestController`. `findById` will simply return null, if the object does not access. But `findById` is accessing the database immediately. See hibernate docs for details :wink:

marcjansen:
I have added some minor comments but really like this. If you agree, and it isn't too much work, you can probably adress them. Either with comments or with actual commits.

Thanks anyhow and feel free to merge this once you are sure this is ready for primetime.

src/shogun2-web/src/main/java/de/terrestris/shogun2/rest/AbstractRestController.java
buehner: I had the same though, but did not evaluate yet...

src/shogun2-web/src/main/java/de/terrestris/shogun2/rest/AbstractRestController.java
buehner: No. Log message should be correct :wink: It is the create case here...

src/shogun2-web/src/main/java/de/terrestris/shogun2/rest/AbstractRestController.java
marcjansen: Same is true below where you ar doing string magic to get the classname for logging. This can be evaluated later,I guess

buehner:
Answer to your question regarding POST/PUT with missing/wrong fields:

* If Jackson cannot deserialize, it will (automagically) return a `400 BAD REQUEST` (and we do not even come into our method)
* If we are violating a database constraint (or the user has no ACL-right for update), we will return a `404 NOT FOUND` (but developers can inspect the LOG for detailed information). Regarding a DB constraint violation, this may be a point for another enhancement! Thx for pointing this out...

src/shogun2-web/src/main/java/de/terrestris/shogun2/rest/AbstractRestController.java
buehner: Currently, we only have `saveOrUpdate` (id==null -> save vs id !== null -> update), but hibernate provides explicit save and update methods... Maybe this is another point for the future.

src/shogun2-web/src/main/java/de/terrestris/shogun2/rest/AbstractRestController.java
marcjansen: Ok, then. I misread the diff somehow. 🙈

src/shogun2-web/src/main/java/de/terrestris/shogun2/rest/AbstractRestController.java
marcjansen: Yeah probably.

marcjansen:
Ahhh and btw: tests would be extra 🍨🍦🍰🍫

src/shogun2-core/shogun2-dao/src/main/java/de/terrestris/shogun2/dao/GenericHibernateDao.java
marcjansen: I need to read that stuff, true. But I fully trust you on this here.

buehner:
For some reason the coverage decreased more than 40 % !?!?!??? I hope that my last commit will fix this. But i dont understand why this did not happen before...

buehner:
My suggestion is: The jacoco config was not correct before 7f3a432 but the system was working on cached resources as nothing has changed in the model module since the introduction of jacoco...

marcjansen:
With regard to the massive coverage decrease: Just compare these three coverages and the files that make up the 100%

* [changed by -2.7 %](https://coveralls.io/builds/4452931) for commit 3422c9f
  * 302 of 462 relevant lines covered (36 files)
* [changed by -46.6 %](https://coveralls.io/builds/4466811) for commit f382c2e
 * 305 of 1423 relevant lines covered (67 files)
* [changed by -2.9 %](https://coveralls.io/builds/4467136) for commit 7f3a432b66c137811afe8aa7a687b7c27c1237d0
  * 305 of 468 relevant lines covered  (37 files)

So whatever has increased the number of files / lines (restructuring of dependencies, s.th. else?!?) was the reason for the decrease.

Anyhow, excluding the models is fine for me. Thanks for the continued work on this.

Please merge at will.

marcjansen:
Enough tests :smile: LGTM, please merge

PR #97, opened 2015-12-16
Reset & update password

dnlkoch:
The logic presented in this PR allows any registered SHOGun2-User to reset their password manually.

This is done by the following workflow:

* A user requests a password-reset by sending their account-email to the interface `resetPassword.action`, e.g.

   `http://localhost:8080/shogun2/user/resetPassword.action?email=user@shogun2.de`

* If the mail really exists, a new `PasswordResetToken` containing a unique token, expire date and user reference is being generated and persisted.

* A special link for updating the password including this token and id will be send to the user, e.g.:

  `http://localhost:8080/shogun2/login/#passwordchange?id={{token-id}}&token={{token}}`

* The user will be redirected to a (client-)page to update their password if they follow the link.

* The client sends a request to the `changePassword.action` with the provided token, token-id and password, e.g.:

   `http://localhost:8080/shogun2/user/changePassword.action?password={{newPassword}}&id={{tokenId}}&token={{token}}`

* If the token is valid (token exists and is not expired) the user will be authorized to change their password and the token will be deleted.

### DIFF
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -138,6 +138,12 @@
         <!-- Java Transaction API -->
         <jta.version>1.1</jta.version>
 
+        <!-- Java Servlet API -->
+        <javax-servlet-api.version>3.1.0</javax-servlet-api.version>
+
+        <!-- Apache HTTP Client -->
+        <apache-httpclient.version>4.5.1</apache-httpclient.version>
+
         <downloadSources>true</downloadSources>
         <downloadJavadocs>true</downloadJavadocs>
     </properties>
@@ -597,6 +603,18 @@
             </dependency>
 
             <dependency>
+                <groupId>org.apache.httpcomponents</groupId>
+                <artifactId>httpclient</artifactId>
+                <version>${apache-httpclient.version}</version>
+            </dependency>
+
+            <dependency>
+                <groupId>javax.servlet</groupId>
+                <artifactId>javax.servlet-api</artifactId>
+                <version>${javax-servlet-api.version}</version>
+            </dependency>
+
+            <dependency>
                 <groupId>javax.mail</groupId>
                 <artifactId>javax.mail-api</artifactId>
                 <version>${javax-mail-api.version}</version>
@@ -620,4 +638,3 @@
     </dependencyManagement>
 
 </project>
-
--- a/src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/token/PasswordResetToken.java
+++ b/src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/token/PasswordResetToken.java
@@ -1,0 +1,76 @@
+package de.terrestris.shogun2.model.token;
+
+import javax.persistence.Entity;
+import javax.persistence.Table;
+
+import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+import org.apache.commons.lang3.builder.ToStringBuilder;
+import org.apache.commons.lang3.builder.ToStringStyle;
+
+
+@Entity
+@Table
+public class PasswordResetToken extends Token {
+
+	/**
+	 *
+	 */
+	private static final long serialVersionUID = 1L;
+
+	/**
+	 * The password token should be valid for 48h.
+	 */
+	private static final int expiration = 48;
+
+	/**
+	 * Constructor
+	 */
+	public PasswordResetToken() {
+		super(expiration);
+	}
+
+	/**
+	 * @see java.lang.Object#hashCode()
+	 *
+	 *      According to
+	 *      http://stackoverflow.com/questions/27581/overriding-equals
+	 *      -and-hashcode-in-java it is recommended only to use getter-methods
+	 *      when using ORM like Hibernate
+	 */
+	@Override
+	public int hashCode() {
+		// two randomly chosen prime numbers
+		return new HashCodeBuilder(17, 37).
+				appendSuper(super.hashCode()).
+				toHashCode();
+	}
+
+	/**
+	 * @see java.lang.Object#equals(java.lang.Object)
+	 *
+	 *      According to
+	 *      http://stackoverflow.com/questions/27581/overriding-equals
+	 *      -and-hashcode-in-java it is recommended only to use getter-methods
+	 *      when using ORM like Hibernate
+	 */
+	@Override
+	public boolean equals(Object obj) {
+		if (!(obj instanceof PasswordResetToken))
+			return false;
+		PasswordResetToken other = (PasswordResetToken) obj;
+
+		return new EqualsBuilder().
+				appendSuper(super.equals(other)).
+				isEquals();
+	}
+
+	/* (non-Javadoc)
+	 * @see java.lang.Object#toString()
+	 */
+	@Override
+	public String toString() {
+		return ToStringBuilder.reflectionToString(this, ToStringStyle.DEFAULT_STYLE);
+	}
+
+}
--- a/src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/token/PasswordResetToken.java
+++ b/src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/token/PasswordResetToken.java
@@ -8,7 +8,13 @@ import org.apache.commons.lang3.builder.HashCodeBuilder;
 import org.apache.commons.lang3.builder.ToStringBuilder;
 import org.apache.commons.lang3.builder.ToStringStyle;
 
-
+/**
+ *
+ * The PasswordResetToken class.
+ *
+ * @author Daniel Koch
+ *
+ */
 @Entity
 @Table
 public class PasswordResetToken extends Token {
--- a/src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/token/Token.java
+++ b/src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/token/Token.java
@@ -122,7 +122,7 @@ public abstract class Token extends PersistentObject {
 	@Override
 	public int hashCode() {
 		// two randomly chosen prime numbers
-		return new HashCodeBuilder(17, 37).
+		return new HashCodeBuilder(17, 29).
 				appendSuper(super.hashCode()).
 				append(getToken()).
 				append(getExpirationDate()).
--- a/src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/token/Token.java
+++ b/src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/token/Token.java
@@ -1,0 +1,163 @@
+package de.terrestris.shogun2.model.token;
+
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.Inheritance;
+import javax.persistence.InheritanceType;
+import javax.persistence.JoinColumn;
+import javax.persistence.OneToOne;
+
+import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+import org.apache.commons.lang3.builder.ToStringBuilder;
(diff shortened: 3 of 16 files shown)


src/pom.xml
```diff
@@ -587,6 +593,18 @@
...
             <dependency>
+                <groupId>org.apache.httpcomponents</groupId>
+                <artifactId>httpclient</artifactId>
+                <version>${apache-httpclient.version}</version>
+            </dependency>
+
+            <dependency>
+                <groupId>javax.servlet</groupId>
+                <artifactId>javax.servlet-api</artifactId>
+                <version>${javax-servlet-api.version}</version>
```
marcjansen: Doesn't this usually have `<scope>provided</scope>`?

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/token/PasswordResetToken.java
```diff
@@ -0,0 +1,82 @@
...
+
+	/**
+	 *
+	 */
+	private static final long serialVersionUID = 1L;
+
+	/**
+	 * The password token should be valid for 48h.
+	 */
+	private static final int expiration = 48;
```
marcjansen: Would `protected int expiration = 48` (and also in the parent class) be a better choice? I am unsure, probably @buehner has something to say.

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/token/Token.java
```diff
@@ -0,0 +1,163 @@
...
+	@OneToOne
+	@JoinColumn(name = "USER_ID")
+	private User user;
+
+	/**
+	 * Constructor
+	 */
+	protected Token(int expiration) {
+		// set the expiration date
+		this.expirationDate = DateTime.now().plusHours(expiration);
```
marcjansen: Why do we have a constructor with `expiration` when this is also set as a class constant?

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/token/Token.java
```diff
@@ -0,0 +1,163 @@
...
+	private final ReadableDateTime expirationDate;
+
+	/**
+	 * The user who has requested the token. Hereby one user can have one
+	 * token and one token can be used by one user (at the same time) only.
+	 */
+	@OneToOne
+	@JoinColumn(name = "USER_ID")
+	private User user;
+
```
marcjansen: The default no argument constructor is missing.

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/token/Token.java
marcjansen: Ahh I see, because the childclass wants to call it with its class constant, right? Maybe @buhner can say sth. wrt this.

src/shogun2-core/shogun2-service/src/main/java/de/terrestris/shogun2/service/PasswordResetTokenService.java
```diff
@@ -0,0 +1,370 @@
...
+	 *
+	 * @param request
+	 * @param email
+	 * @return
+	 * @throws UsernameNotFoundException
+	 * @throws Exception
+	 * @throws URISyntaxException
+	 */
+	public Boolean sendResetPasswordMail(HttpServletRequest request, String email) throws
+			UsernameNotFoundException, Exception, URISyntaxException {
```
marcjansen: If possible remobve the `Exception`. I don't like the generic type… others may disagree

src/shogun2-core/shogun2-service/src/main/java/de/terrestris/shogun2/service/PasswordResetTokenService.java
```diff
@@ -0,0 +1,370 @@
...
+	/**
+	 *
+	 * @param password
+	 * @param id
+	 * @param token
+	 * @return
+	 * @throws Exception
+	 */
+	public Boolean changePassword(String password, int id, String token)
+			throws Exception {
```
marcjansen: Same here. I would like to have more specific exceptions, if possible.

src/shogun2-core/shogun2-service/src/main/java/de/terrestris/shogun2/service/PasswordResetTokenService.java
```diff
@@ -0,0 +1,370 @@
...
+			user.setPassword(passwordEncoder.encode(password));
+			userService.updateExistingUser(user);
+			LOG.debug("Successfully updated the password.");
+		} catch(Exception e) {
+			throw new Exception("Could not update the password: "
+					+ e.getMessage());
+		}
+
+		// delete the token
+		dao.delete(passwordResetToken);
```
marcjansen: Should this be moved to the above try-catch in a finally? When there was a problem with the token, we simply throw and the token will never be removed, right?

src/shogun2-core/shogun2-service/src/main/java/de/terrestris/shogun2/service/PasswordResetTokenService.java
```diff
@@ -0,0 +1,370 @@
...
+
+		// get the webapp URI
+		URI appURI = contextUtil.getApplicationURIFromRequest(request);
+
+		// build the change-password URI send to the user
+		URI tokenURI = new URIBuilder()
+				.setScheme(appURI.getScheme())
+				.setHost(appURI.getHost())
+				.setPort(appURI.getPort())
+				.setPath(appURI.getPath() + changePasswordPath)
```
marcjansen: How is the autowired `changePasswordPath` supposed to look like? Do I need a leading slash?

src/shogun2-core/shogun2-service/src/main/java/de/terrestris/shogun2/service/PasswordResetTokenService.java
```diff
@@ -0,0 +1,370 @@
...
+				.setParameter("id", String.valueOf(resetPasswordToken.getId()))
+				.setParameter("token", resetPasswordToken.getToken())
+				.build();
```
marcjansen: This is a nice way of constructing the URI. A future PR may refactor this into a reusable utility class, as I think we'll need this in many places.

src/shogun2-core/shogun2-service/src/main/java/de/terrestris/shogun2/service/UserService.java
```diff
@@ -38,6 +48,20 @@ public User findByAccountName(String accountName) {
...
 	/**
+	 *
+	 * @param email
+	 * @return
+	 */
+	public User findByEmail(String email) {
+
+		SimpleExpression eqEmail = Restrictions.eq("email",
+				email);
+		User user = dao.findByUniqueCriteria(eqEmail);
```
marcjansen: Do we already have a unique constraint against `email`? I *think* we may want one.

src/shogun2-core/shogun2-service/src/main/java/de/terrestris/shogun2/service/UserService.java
```diff
@@ -66,6 +90,24 @@ public User createNewUser(User user, boolean encryptPassword) {
...
+
+		if(user.getId() == null) {
+			// to be sure that we are in the
+			// "update" case, the id must not be null
+			return user;
+		}
+
+		dao.saveOrUpdate(user);
+
+		return user;
```
marcjansen: Shoul we return the returned `user`? It has changed, so the passed user may not be accurate anymore.

src/shogun2-security/src/test/resources/META-INF/spring/test-context-acl.xml
```diff
@@ -19,7 +19,12 @@
     </jdbc:initialize-database>
 
     <context:component-scan base-package="de.terrestris.shogun2.security,
-        de.terrestris.shogun2.service, de.terrestris.shogun2.dao" />
+            de.terrestris.shogun2.dao" />
+
+    <context:component-scan base-package="de.terrestris.shogun2.service" use-default-filters="false">
+        <context:include-filter type="aspectj" expression="de.terrestris.shogun2.service..ApplicationService*"/>
+        <context:include-filter type="aspectj" expression="de.terrestris.shogun2.service..PersonService*"/>
+    </context:component-scan>
```
marcjansen: Just curious, what do the `..` mean?

src/shogun2-util/pom.xml
```diff
@@ -74,6 +74,16 @@
             <artifactId>greenmail</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>javax.servlet</groupId>
+            <artifactId>javax.servlet-api</artifactId>
```
marcjansen: See above, probably mneeds scope = provided

marcjansen:
This looks good @dnlkoch. I have added some comments and would value the feedback of @buehner for certain aspects.

The comments with regard to `Exception` should not block this, as we need general consensus for this thing.

I am :+1: on merging, but would love to heart @buehner's thoughts first.

Thanks anyhow!

marcjansen:
@buehner may also be able to help with some tests, so that coveralls is happy.

src/pom.xml
buehner: Thats a good point! We should try it out...

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/token/PasswordResetToken.java
buehner: Yes. I'm a bit confused as we have two static variables with different values!??

`private static final Token.expiration` and `private static final PasswordResetToken.expiration` ?

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/token/Token.java
buehner: :+1:

src/shogun2-core/shogun2-service/src/main/java/de/terrestris/shogun2/service/PasswordResetTokenService.java
buehner: :+1:

src/shogun2-core/shogun2-service/src/main/java/de/terrestris/shogun2/service/PasswordResetTokenService.java
buehner: :+1:

src/shogun2-core/shogun2-service/src/main/java/de/terrestris/shogun2/service/UserService.java
buehner: You're right. There is currently NO `@Column(unique = true)` on the email property. This has to be done in the `Person` model!

src/shogun2-core/shogun2-service/src/main/java/de/terrestris/shogun2/service/UserService.java
buehner: It is possible that this will even work this way, but it agree that it would be nicer to explicitly return the saved user.

src/shogun2-security/src/test/resources/META-INF/spring/test-context-acl.xml
buehner: The two dots match any number of arguments. It is an aspectj expression

buehner:
I rebased this PR/branch to resolve the conflicts with master. When CI is finished, i will merge and open a follow up PR to address the open issues.

src/pom.xml
buehner: The answer is: We **can't** use `provided` here as we are explicitly using `javax.servlet.http.HttpServletRequest` outside of any servlet container (e.g. in the `util`) module.

buehner:
There is a (WIP) follow up PR: #101

Issue #93, opened 2015-12-14
Creation of table "user" fails with PostgreSQL


chrismayer:
The creation of the database table "user" on startup of the webapp fails with PostgreSQL. The Reason is the fact that "user" is a reserved word in PostgreSQL (see [here](http://www.postgresql.org/docs/9.4/static/sql-keywords-appendix.html)).
To solve this the table name has to be enclosed by double quotes, which can be achieved in javax.persistance by adding an escaped table name declaration to the model, like this 

```java
@Table(name="\"user\"")
```
If you confirm this error and go along with the proposed solution, I'll provide a PR.

marcjansen:
Thanks for the bug-report, @chrismayer. I can live with any solution, but would love to have sth. like e.g. a configurable prefix for tables, possibly in a `.properties`-file.

I don't know if this is doable, but that would make this complete DDL-stuff less error-prone.

chrismayer:
Thanks for your reply @marcjansen. I agree that prefixing the table names is a very common practice to make this DDL stuff less error-prone. It seems that there is a hibernate way of doing this with a custom NamingStrategy implementation, as suggested [here](http://stackoverflow.com/questions/4313095/jpa-hibernate-and-custom-table-prefixes) or shown [here](http://www.petrikainulainen.net/programming/tips-and-tricks/implementing-a-custom-namingstrategy-with-hibernate/).

marcjansen:
Yeah this sounds great. Is there any chance we may at some time support sth. like:

```java
@Table(name="${common_tbl_prefix}user")
```

I'd value @buehner feedback here.

marcjansen:
I am plus one on a custom naming strategy as outlined here: http://www.petrikainulainen.net/programming/tips-and-tricks/implementing-a-custom-namingstrategy-with-hibernate/

Inside that naming startegy we can outo wire a bean containg the configurable prefix.

I want to inherit as much of the parent class as possible.

chrismayer:
I am also fine with a custom naming strategy adding a configurable prefix to the table names.

Anyone else an opinion on that?

buehner:
I like the custom naming strategy approach and could/would give it a try within the next days (or even hours).

What's your opinion regarding the requirements?

As a first solution, i would suggest to implement a strategy that

* generates table names in **plural** like in the example above, but more powerful by respecting some basic rules: http://www.edufind.com/english-grammar/plural-nouns/
* (Optionally) allows the use of a prefix for all table names

[...]

PR #99, opened 2015-12-18
Update dependencies

buehner:
I would like to update the versions of the dependencies for the following reasons:

* I had problems downloading the sources/javadocs for hibernate (which seems to work with the new version)
* I need these sources/docs for an easier development/fix of the naming strategy (see #93 )

*Note:* I had to add a new dependency to the DAO module (Java Transaction API). Otherwise the DAO tests failed. It seems that this is related to hibernate, but i don't know yet why this is not mentioned in the hibernate release notes.

### DIFF
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -84,16 +84,16 @@
         <tomcat.version>8.0.28</tomcat.version>
 
         <!-- Core -->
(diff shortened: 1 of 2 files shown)


marcjansen:
Good to go. Thanks @buehner. Please merge.

buehner:
:+1:

PR #100, opened 2015-12-21
Introduce Hibernate naming strategies

buehner:
Regarding #93 , this PR introduces a DDL naming strategy for SHOGun2:

Since Hibernate 5, the naming strategy is split up in two parts:

* *ImplicitNamingStrategy* - If an explicit naming configuration is given via `@Table` (or `@JoinTable`), this strategy will **NOT** be executed. Otherwise: Based on the class name, an implicit name for the entity will be created. In our case, this strategy will produce *plural forms* of the singular class name forms, e.g. `Application -> Applications`. 
* *PhysicalNamingStrategy* - This strategy will **always** be executed: Based on the result of the implicit naming strategy or an explicit naming configuration, this strategy will create a physical name for tables/schemas etc.  In our case, the implementation will
  1. transform all entity names to **lower case**
  2. **limit the length of an identifier** according to the given JDBC context (currently H2, Oracle and PostgreSQL are supported)

As a result of this implementation, a lot of `@Table` and `@JoinTable` annotations have been cleaned up. Even though, in case of most `@JoinTable`s the naming should still be done manually to avoid that the physical naming strategy cuts off the (implicitly) generated join table names (including `_`) in case of a limitation.

Tests are included!


marcjansen:
This is looking very neat, @buehner!

How do you configure a prefix for all tables, be they explicitly named or generated from a model? This is missing, right?

I imagen that an application can create a bean / instance of `ImplicitNamingStrategyShogun2` and give it the prefix to use.

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/User.java
```diff
@@ -43,7 +43,6 @@
 
 	@ManyToMany
 	@JoinTable(
-		name = "USER_ROLE",
```
marcjansen: This will become `users_roles`, right?

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/map/MapControl.java
```diff
@@ -50,15 +50,15 @@
 	 *
 	 */
 	@ElementCollection
-	@MapKeyColumn(name = "MAPCTRLPROPERTY")
+	@MapKeyColumn(name = "PROPERTY")
 	@Column(name = "VALUE")
 	@CollectionTable(
-			name = "MAPCONTROL_MAPCTRLPROPERTIES",
-			joinColumns = @JoinColumn(name = "MAPCONTROL_ID")
+		name = "MAPCONTROLS_PROPERTIES",
```
marcjansen: Why do you explicitly give a name here, bit not everywhere else?

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/layer/source/XyzLayerDataSource.java
```diff
@@ -43,7 +43,7 @@
 	private Extent extent;
 
 	@OneToMany
-	@JoinTable(name = "LAYERDATASOURCE_RESOLUTIONS")
+	@JoinTable(name = "XYZLAYERDATASRC_RESOLUTIONS")
```
marcjansen: Why do you explicitly give a name here, but not everywhere else?

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/layer/source/TileWmsLayerDataSource.java
```diff
@@ -63,9 +63,9 @@
 
 	@ManyToMany
 	@JoinTable(
-			name = "TILEWMSLAYERDATASOURCE_STYLE",
-			joinColumns = { @JoinColumn(name = "TILEWMSLAYERDATASOURCE_ID") },
-			inverseJoinColumns = { @JoinColumn(name = "STYLE_ID") }
+		name = "TILEWMSLAYERDATASOURCE_STYLE",
```
marcjansen: Why do you explicitly give a name here, but not everywhere else?

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/layer/source/TileWmsLayerDataSource.java
```diff
@@ -46,9 +46,9 @@
 
 	@ManyToMany
 	@JoinTable(
-			name = "TILEWMSLAYERDATASRC_LAYERNAME",
-			joinColumns = { @JoinColumn(name = "TILEWMSLAYERDATASOURCE_ID") },
-			inverseJoinColumns = { @JoinColumn(name = "LAYERNAME_ID") }
+		name = "TILEWMSLAYERDATASRC_LAYERNAME",
```
marcjansen: Why do you explicitly give a name here, but not everywhere else?

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/layer/source/ImageWmsLayerDataSource.java
```diff
@@ -44,34 +44,34 @@
...
 	)
 	@JsonIdentityReference(alwaysAsId = true)
 	private List<GeoWebServiceLayerName> layerNames;
 
 	@ManyToMany
 	@JoinTable(
-			name = "IMAGEWMSLAYERDATASOURCE_STYLE",
-			joinColumns = { @JoinColumn(name = "IMAGEWMSLAYERDATASOURCE_ID") },
-			inverseJoinColumns = { @JoinColumn(name = "STYLE_ID") }
+		name = "IMAGEWMSLAYERDATASOURCE_STYLE",
```
marcjansen: Why do you explicitly give a name here, but not everywhere else?

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/module/CompositeModule.java
```diff
@@ -54,9 +54,9 @@
 	@ManyToMany
 	@Cascade(CascadeType.SAVE_UPDATE)
 	@JoinTable(
-			name = "MODULE_SUBMODULE",
-			joinColumns = { @JoinColumn(name = "MODULE_ID") },
-			inverseJoinColumns = { @JoinColumn(name = "SUBMODULE_ID") }
+		name = "MODULE_SUBMODULE",
```
marcjansen: Why do you explicitly give a name here, but not everywhere else? Also shouldn't this be `MODULES_SUBMODULES``

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/module/OverviewMap.java
```diff
@@ -54,11 +54,10 @@
 	/**
 	 * The layers used within this OverviewMap.
 	 */
-	@ElementCollection
+	@ManyToMany
```
marcjansen: Why is this change needed?

src/shogun2-util/src/main/java/de/terrestris/shogun2/util/naming/ImplicitNamingStrategyShogun2.java
```diff
@@ -0,0 +1,76 @@
...
+ */
+public class ImplicitNamingStrategyShogun2 extends ImplicitNamingStrategyJpaCompliantImpl {
+
+	private static final long serialVersionUID = 1L;
+
+	private static final String PLURAL_SUFFIX_S = "s";
+
+	private static final String LAST_CHAR_S = "s";
+	private static final String LAST_CHAR_X = "x";
+	private static final String LAST_CHAR_Z = "z";
```
marcjansen: Shoukld the above four constants probably be of type `char`?

src/shogun2-util/src/main/java/de/terrestris/shogun2/util/naming/ImplicitNamingStrategyShogun2.java
```diff
@@ -0,0 +1,76 @@
...
+	private static final String LAST_CHARS_CH = "ch";
+	private static final String LAST_CHARS_SH = "sh";
+	private static final String PLURAL_SUFFIX_ES = "es";
+
+	private static final String LAST_CHAR_Y = "y";
```
marcjansen: Should this be type `char`?

src/shogun2-util/src/main/java/de/terrestris/shogun2/util/naming/ImplicitNamingStrategyShogun2.java
```diff
@@ -0,0 +1,76 @@
...
+	 * @return
+	 */
+	private String transformToPluralForm(String singular) {
+		StringBuilder plural = new StringBuilder();
+
+		// start with singular form
+		plural.append(singular);
+
+		if (singular.endsWith(LAST_CHAR_Y)) {
+			// replace last char with suffix form
```
marcjansen: Perhaps an inline example helps future readers:

```java
// e.g. "city" => "cities"
```

src/shogun2-util/src/main/java/de/terrestris/shogun2/util/naming/ImplicitNamingStrategyShogun2.java
```diff
@@ -0,0 +1,76 @@
...
+			int length = plural.length();
+			plural.replace(length - 1, length, PLURAL_SUFFIX_IES);
+
+		} else if (singular.endsWith(LAST_CHAR_S)
+				|| singular.endsWith(LAST_CHAR_X)
+				|| singular.endsWith(LAST_CHAR_Z)
+				|| singular.endsWith(LAST_CHARS_CH)
+				|| singular.endsWith(LAST_CHARS_SH)) {
+
```
marcjansen: This could also benefit from inilne examples (takle the values from the tests)

marcjansen:
I have added some minor inline comments and on top of the inline diffs some other qquestions. Thanks for the work on this.

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/User.java
buehner: Correct!

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/map/MapControl.java
buehner: Because the implicitly generated value in this case would exceed 30 chars (but the PhysicalNamingStrategy cuts off after 30 chars in case of oracle)

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/layer/source/XyzLayerDataSource.java
buehner: see above...

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/layer/source/TileWmsLayerDataSource.java
buehner: see above...

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/layer/source/TileWmsLayerDataSource.java
buehner: see above...

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/layer/source/ImageWmsLayerDataSource.java
buehner: see above...

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/module/CompositeModule.java
buehner: Because `MODULES_SUBMODULES` is nicer than `COMPOSITEMODULES_MODULES` ;-) I will correct this to the plural form in a follow up commit

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/module/OverviewMap.java
buehner: `@ElementCollection` should only be used for basic or embedabble types, which was not correct here. It seems that not all of the config was respected in case of `@ElementCollection`

src/shogun2-util/src/main/java/de/terrestris/shogun2/util/naming/ImplicitNamingStrategyShogun2.java
buehner: Why not!? I will change this in a follow up commit

src/shogun2-util/src/main/java/de/terrestris/shogun2/util/naming/ImplicitNamingStrategyShogun2.java
buehner: see above

src/shogun2-util/src/main/java/de/terrestris/shogun2/util/naming/ImplicitNamingStrategyShogun2.java
buehner: Good point! I will add some inline comments

src/shogun2-util/src/main/java/de/terrestris/shogun2/util/naming/ImplicitNamingStrategyShogun2.java
buehner: see above

buehner:
To answer your questions, @marcjansen :

* You're right. I simply forgot the prefix. I will add this in a follow up commit
* Currently, our strategy only applies to names of tables/join tables.
* Here you can see the possibilities we have regarding the implicit strategy: https://docs.jboss.org/hibernate/orm/5.0/javadocs/org/hibernate/boot/model/naming/ImplicitNamingStrategy.html
* Here the doc for the physical naming strategy: https://docs.jboss.org/hibernate/orm/5.0/javadocs/org/hibernate/boot/model/naming/PhysicalNamingStrategy.html

Regarding your comments, i will add some commits soon. Thx!

buehner:
I pushed some new commits that address your notes. Most important change: columns will now also be limited and lower case!

Feel free to merge, if you are happy. Thx

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/User.java
marcjansen: OK. then I understood the concept :-)

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/module/OverviewMap.java
marcjansen: OK

src/shogun2-util/src/main/java/de/terrestris/shogun2/util/naming/PhysicalNamingStrategyShogun2.java
```diff
@@ -0,0 +1,131 @@
...
+
+	private static final long serialVersionUID = 1L;
+
+	protected static final int LENGTH_LIMIT_ORACLE = 30;
+
+	protected static final int LENGTH_LIMIT_POSTGRESQL = 63;
+
+	@Autowired(required = false)
+	@Qualifier("tablePrefix")
+	private String tablePrefix;
```
marcjansen: Why did you chose

```java
private String tablePrefix;
```

over

```java
private String tablePrefix = null;
```

or even

```java
private String tablePrefix = "";
```

are these equivalent?

marcjansen:
I have added some more comments. Can you pointme to a test where the prefix behaviour is tested?

Also (but probably better done in a follow-up PR) I would like to limit the prefix to max 4 characters, with only a subset of allowe chars: `/^[a-z_]{0,4}/$`. What do you think?

src/shogun2-util/src/main/java/de/terrestris/shogun2/util/naming/PhysicalNamingStrategyShogun2.java
buehner: At first, the `@Autowired` annotation is configured with `required=false`, i.e. the prefix value is not mandatory.

`private String tablePrefix;` is the equal to `private String tablePrefix = null;` as `null` is the default value.

On top of that, the prefix value will only be used if it is not null:

https://github.com/buehner/shogun2/blob/naming-strategy/src/shogun2-util/src/main/java/de/terrestris/shogun2/util/naming/PhysicalNamingStrategyShogun2.java#L69-L71

buehner:
@marcjansen I added three tests for the table prefix in 82a1b42

Please review and merge. Thx

marcjansen:
Thanks. Good to go.

PR #101, opened 2015-12-23
Refine password reset

buehner:
This is a follow up PR for #97

### DIFF
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -115,7 +115,7 @@
         <hamcrest.version>1.3</hamcrest.version>
         <mockito.version>1.10.19</mockito.version>
         <jsonpath.version>1.2.0</jsonpath.version>
-        <greenmail.version>1.4.1</greenmail.version>
+        <greenmail.version>1.4.2-SNAPSHOT</greenmail.version>
 
         <!-- AOP -->
         <aspectj.version>1.8.7</aspectj.version>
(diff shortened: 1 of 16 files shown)


buehner:
The failing test seems to be related to: https://github.com/greenmail-mail-test/greenmail/issues/76

buehner:
d6cd74d is mainly a refactoring of the two `Token` classes:

* The (abstract) `Token` class is now fully responsible for the creation of the following `final` values on instantiation:
  * token value (UUID)
  * expirationDate (which is now computed based on a `expiration` value (only passed via constructor, but removed as a property) and the `created` property, i.e. `Date.now()`)
* The one-to-one `User`-relation has been moved from `Token` to `PasswordResetToken` as this relation is what distinguishes the child class from the `Token` class...
* The `Token` class now has a nice helper method: `expiresWithin(int minutes) : boolean`

On top of that, the `PasswordResetTokenService` has been adapted. The main difference is that an existing token for a user will only be deleted if it expires within the next 5 minutes. Otherwise the existing one can still be used by the user.

Maybe we should think a bit more about the `Token` class: Is it really wanted that this class is `abstract`? Maybe we want to use a simple `Token` instance in future for other reasons...

marcjansen:
I suggest to postpone a review / merge of this once we can discuss it with @dnlkoch, as there may be client projects depending on this.

Nice work btw.

buehner:
@dnlkoch Please review now! Missing tests will be added in another PR

dnlkoch:
Good work! I like your additions and will merge now.

PR #102, opened 2016-01-04
Introduces account state handling.

KaiVolland:


src/shogun2-security/src/test/java/de/terrestris/shogun2/security/Shogun2AuthenticationProviderTest.java
```diff
@@ -195,6 +202,31 @@ public void authenticate_shouldThrowBadCredentialsExceptionIfPasswordDoesNotMatc
...
+	 * active.
+	 */
+	@Test(expected=DisabledException.class)
+	public void authenticate_shouldThrowDisabledExceptionIfUserIsInactive() {
+
+		// 1. Mock an authentication request object
+		final String shogun2UserName = "user";
+		final String correctPassword = "correctPassword";
+		final User userToAuth = createUserMock(shogun2UserName, correctPassword);
+
```
buehner: Please set the `active` flag explicitly to `false` (although it is the default value).

Possibly it makes sense to refactor the `createUserMock` method by adding an additional parameter `active`...!?

dnlkoch:
:+1:

PR #105, opened 2016-01-07
Rename update User By Session Interface

KaiVolland:
The SecurityContextHolder holds a static copy of the user from the moment he logged in. So we need to get the current instance from dao.


buehner:
:+1:

PR #106, opened 2016-01-08
Embed the AWT 2D Points to x and y columns

### LINKED ISSUE #94: Property "center" of MapConfig transformed to bytea in PostgreSQL
When setting up SHOGun 2 with PostgreSQL (v9.4) as database I encountered that the property "center" of the MapConfig model is transformed to the field type "bytea" in the database:

mapconfig-model

Is this a wanted behaviour?

buehner:
This will avoid that hibernate uses BLOB columns for the Java AWT 2D point data. By using the `@Embedded` annotation in combination with `@AttributeOverrides`, the `x` and `y` values will now be stored in different columns with a numeric type, which makes it easier to debug/develop or change data in the database.

Closes #94.


marcjansen:
Superb! Please merge!

buehner:
:+1:

PR #108, opened 2016-01-08
Add a BeanUtil to copy only non-null properties

buehner:
Based on http://stackoverflow.com/a/3521314, this new `NullAwareBeanUtilsBean` allows the use of `copyProperties()` to copy only those properties that are NOT null.

### DIFF
--- a/src/shogun2-util/pom.xml
+++ b/src/shogun2-util/pom.xml
@@ -90,6 +90,12 @@
             <artifactId>httpclient</artifactId>
         </dependency>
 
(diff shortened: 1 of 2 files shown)


weskamm:
Thanks for the nice addition, will merge

marcjansen:
Which version of this dependency will be used?

buehner:
@marcjansen The one defined in https://github.com/terrestris/shogun2/blob/master/src/pom.xml, which is currently version 1.9.2

PR #109, opened 2016-01-11
Added JSONProperty Annotation for better password handling

weskamm:
The JSONProperty Annotation is now used for the users password in order to avoid sending the password from backend to frontend but allow the sending from frontend to backend (e.g. in registration process / login)


buehner:
LGTM, thx!

PR #110, opened 2016-01-11
Use persisted role

buehner:
* Add a role service to access a persisted role by name (tests included)
* Adapt user service to use persisted role when activating an account: The previous code was only working if the system booted with hibernate/CREATE mode and SHOGun2 init content flag set to true. In other cases, the autowired defaultUserRole had no ID


weskamm:
lgtm, merging...

PR #111, opened 2016-01-11
Add token service tests

buehner:
Title says it all


marcjansen:
Do we need to change the visibility of `protected E getValidTokenForUser()` as well?

weskamm:
Thanks for the changes and additional tests, will merge

buehner:
@marcjansen Not currently as we don't need it at the moment. We can change this at a later point if we should really need it

PR #112, opened 2016-01-11
Refactor user registration

buehner:
Refactor the user registration interface to accept a `User` object, but not single properties (email and password). By doing, this we can easily use the SHOGun registration interface with user objects that extend the SHOGun2 `User` class


weskamm:
Thanks again, helpful for project specific handling, will merge...

PR #114, opened 2016-01-13
Make DAOs, services and web controller extendible

buehner:
This is a major PR that makes all DAOs, services and web controllers extendible. This change is important to avoid code duplication in this project (or projects that extend SHOGun2). By using Java Generics, components like DAOs or services in subclasses will always have the correct type (and thereby provide the full API for their type), which is very nice because developers are not forced anymore to cast these components in certain situations.

Long story short: Defining a new (sub-)class (using Java Generics) is more complex than before, but using/developing the API will be more fun now.

On top of that one minor change has been introduced: Abstract parent classes (i.e. `GenericHibernateDao`, `AbstractDaoService` and `AbstractWebController`) now provide a LOGGER instance that can easily be used in all their subclasses, i.e. subclasses don't need to define their logger instance anymore as the Logger in the above mentioned parent classes is defined as:

``` java
protected final Logger LOG = Logger.getLogger(getClass());
```

where `getClass()` will always return the type of the concrete subclass...


src/shogun2-core/shogun2-dao/src/main/java/de/terrestris/shogun2/dao/AbstractTokenDao.java
```diff
@@ -0,0 +1,27 @@
+package de.terrestris.shogun2.dao;
+
+import org.springframework.stereotype.Repository;
+
+import de.terrestris.shogun2.model.token.Token;
+
+/**
+ * As the {@link Token} class is abstract, this class will also be abstract.
+ * There will also be NO {@link Repository} annotation here.
```
marcjansen: So the import is obnly ther to link to the not-used annotation? Can't we simpy use the text `org.springframework.stereotype.Repository`?

buehner: Haha. You're right. The import is only there because `Repository` is mentioned in the JavaDoc. If you want the `Repository` being linked in the JavaDoc, we should not change this. ;-)

buehner:
Please wait with merging. I forgot the changes in the archetype... I will add some commits soon.

src/shogun2-core/shogun2-service/src/main/java/de/terrestris/shogun2/service/PersonService.java
```diff
@@ -1,17 +1,36 @@
...
+ *
  */
 @Service("personService")
-public class PersonService extends AbstractExtDirectCrudService<Person> {
+public class PersonService<E extends Person, D extends PersonDao<E>> extends
+		AbstractExtDirectCrudService<E, D> {
 
+	E peter() {
+		return null;
+	}
```
marcjansen: :lollipop: :lollipop: :lollipop:

buehner: Oooops :cold_sweat:

I will remove this in a following commit ;-)

src/shogun2-core/shogun2-dao/src/main/java/de/terrestris/shogun2/dao/AbstractTokenDao.java
marcjansen: I don't really care. Was just surprised at first.

marcjansen:
Great and impressive work. Very nice that there were nearly no changes to the tests involved. Do we have a readme for creating the sub-classes? Might well be worth adding one.

Please merge once the minor comments are adressed, the archetype is updated, and the lights a green again.

:clap:

buehner:
We don't have a README that explains how to extend existing classes. But the result of the archetype contains demos for Model/DAO/Service/Controller...

marcjansen:
> … the result of the archetype contains demos for Model/DAO/Service/Controller

OK for me.

buehner:
This can be merged now!

PR #115, opened 2016-01-13
Add SHOGun2 logo

dnlkoch:
Here we are!


marcjansen:
So nice! Good work from the two of us :trollface:

Please merge!

PR #120, opened 2016-01-14
Archetype releasing

buehner:
This PR adapts the release script and pom.xml to set the correct SHOGun2 version in the archetype resources that are not respected by the maven-release-plugin.


dnlkoch:
gif

marcjansen:
Very elegant solution!

PR #121, opened 2016-01-14
Assure compatablity with Java 1.7

buehner:
Title says it all.

Until now, the `maven-compiler-plugin` was using the current Java version of the system. By setting the target version to 1.7, we can be sure that all artifacts are always compatible with version 1.7 - disregarding which version was used to compile (e.g. version 1.8)


dnlkoch:
Very useful! I'll merge.

PR #123, opened 2016-01-14
Make members available in subclasses

buehner:
Title says it all...


marcjansen:
Good to go, thanks.

We may want to revisit other classes whether they also have members which aren't visible in subclasses.

PR #124, opened 2016-01-14
Properly handle null when creating result sets

marcjansen:
Title says it all.


buehner:
:+1: Nice catch @marcjansen 

Thx for adding the tests. I will merge now

PR #125, opened 2016-01-15
Provide runtime entity class for controllers

buehner:
This PR forces all web controller classes to provide their entity class by calling an appropriate constructor. This is very helpful when extending a web controller to create "generic" instances during runtime in a parent class (see change in UserController).

This pattern is already implemented in the DAO layer. See http://stackoverflow.com/a/3403987 for details.

I will also introduce this pattern for the service layer in a following PR


KaiVolland:
LGTM

PR #128, opened 2016-01-26
Refactoring of the REST-Update method.

KaiVolland:
This PR addresses the REST-PUT interface.

- [x] The PUT-Interface should be able to work with partially changes of an object.

Application sampledata 1:
```js
{
  "id": 57,
  "name": "Peter"
}
```

### DIFF
--- a/src/shogun2-core/shogun2-dao/src/main/java/de/terrestris/shogun2/dao/ExtentDao.java
+++ b/src/shogun2-core/shogun2-dao/src/main/java/de/terrestris/shogun2/dao/ExtentDao.java
@@ -1,0 +1,28 @@
+package de.terrestris.shogun2.dao;
+
+import org.springframework.stereotype.Repository;
+
+import de.terrestris.shogun2.model.layer.util.Extent;
+
+@Repository("extentDao")
+public class ExtentDao<E extends Extent> extends
+		GenericHibernateDao<E, Integer> {
+
+	/**
+	 * Public default constructor for this DAO.
+	 */
+	@SuppressWarnings("unchecked")
+	public ExtentDao() {
+		super((Class<E>) Extent.class);
+	}
+
+	/**
+	 * Constructor that has to be called by subclasses.
+	 *
+	 * @param clazz
+	 */
+	protected ExtentDao(Class<E> clazz) {
+		super(clazz);
+	}
+
+}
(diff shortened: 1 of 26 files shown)


buehner:
The solution for partial updates on the top level seems very nice to me!

But i am not sure if we really want to allow the updating of nested objects/collections as this would (currently) only work if the nested object is annotated with `@Cascade(CascadeType.SAVE_UPDATE)`. As we are talking about the `AbstractRestController` here, we cannot assume that all nested objects (of arbitrary entities) are annotated like that.

To update the viewport of an application, wouldn't it be a more RESTful approach to `PUT` 
```
"viewport": {
    "id:" 56,
    "name": "Silke"
}
```
against something like
```
http://localhost:8080/{{shogun2-app}}/rest/applications/{appId}/viewport
```
or
```
http://localhost:8080/{{shogun2-app}}/rest/applications/{appId}/viewport/{{vpId}}
```

This would lead to an implementation in the `ApplicationRestController`. Client applications would then possibly have to fire multiple PUT requests...

buehner:
I just added an enhancement of the current state and fixed the tests

KaiVolland:
> But i am not sure if we really want to allow the updating of nested objects/collections as this would (currently) only work if the nested object is annotated with @Cascade(CascadeType.SAVE_UPDATE)

We allready have this annotation on nearly every property e.g.:

https://github.com/terrestris/shogun2/blob/master/src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/Application.java#L82-L83

https://github.com/terrestris/shogun2/blob/master/src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/map/MapConfig.java#L70-L71

I would appreciate to be able to update nested objects/collections with only one PUT request. We could add errorhandling if the annotation on an edited property is missing.

The adavantage would be the you could easily update applications from the Client side.
This might be an disadavantag too, as you could even easily send a malformed json the corrupts your whole data.

Would be cool if some other devs could submit there opinion on this. @marcjansen @weskamm @dnlkoch @chrismayer

buehner:
> We allready have this annotation on nearly every property

Yes, but these annotations are only there because it was easy and fast (-> saved code) when we were developing the initialization service. I dont think that it is best practice (e.g. regarding performance) and the Cascade annotations should only be used/configured if we have a good reason for that. (Which means on the other side that these annotations should be removed from the code someday)

Its a hard topic and i dont have a strong meaning on this. But if you really want to update all nested elements, i dont think that we will get this running with the changes contained in this PR

KaiVolland:
I'm done from my side. Any more comments? @buehner

weskamm:
As i dont see any conflicts here and as far as i know its already working fine in our latest project, i am going to merge this one..

PR #130, opened 2016-02-05
Remove ACL and use custom PermissionEvaluator

buehner:
This is a major PR that completely removes the ACL stuff and introduces a custom PermissionEvaluator to handle the access rights for domain objects.

**Why?** ACL is complex to handle and usually used by large companies (but not small projects like SHOGun2). Implementing custom permission evaluators makes it much easier for developers to get into that stuff.

Major changes in this PR:

* Introduce a new model class `SecuredPersistentObject` that contains maps with user/group permissions
  * All entities that should be secured on the domain/object level (for now only `Application` and `Person/User`) should extend this class
* Introduce a `PermissionEvaluator` for SHOGun2
* Introduce a `ConfigHolder` object that allows us to use property values in annotations. We now can configure ROLE names in property files. (See `AbstractCrudService`)

**TODO:**

* Write tests for new code


weskamm:
I checked this PR against our current SHOGun2 project and after rebasing this against master and adapting the changes (mainly property files) it works really fine!
So +1 from me after this has been rebased.

buehner:
I added some tests. Feel free to merge.

weskamm:
Nice one, thanks, will merge..

PR #131, opened 2016-02-12
Minor cleanup

buehner:
* Remove unused imports
* Remove unused set-beans
  * These beans have never been used as the `@Autowired` annotations for the sets in the `ContentInitializer` (e.g. `@Autowired(required = false) private Set<User> defaultUsers;`) will automatically inject all available beans of the corresponding type.


KaiVolland:
Nice. Cleanups are allways welcome. Once the test are green merge it.

PR #113, opened 2016-01-11
Towards an automated (Central Repository) deploy process

dnlkoch:
This PR suggests an automated workflow for deploying the current (SNAPSHOT and release) state of the `shogun2` repository to the *Central Repository* using the `maven-release-plugin`. The presented steps are mainly inspired by the official guide as described [here](http://central.sonatype.org/pages/ossrh-guide.html).

The deployment process has to be differentiated between a *full* and *SNAPSHOT* release:

  1. A *full* release will happen only after a manual execution of the provided script `release.sh` (including the wished version as only parameter, e.g. `release.sh 0.0.1`). Running this script will do the following steps for you (with the help of the `maven-release-plugin`):

  > * Check that there are no uncommitted changes in the sources
  > * Check that there are no SNAPSHOT dependencies
  > * Change the version in the POMs from x-SNAPSHOT to a new version (you will be prompted for   the versions to use)
  > * Transform the SCM information in the POM to include the final destination of the tag
  > * Run the project tests against the modified POMs to confirm everything is in working order
  > * Commit the modified POMs
  > * Tag the code in the SCM with a version name (this will be prompted for)
  > * Bump the version in the POMs to a new value y-SNAPSHOT (these values will also be prompted for)
  > * Commit the modified POMs
  
  ([Source](http://maven.apache.org/maven-release/maven-release-plugin/examples/prepare-release.html))

  > * Checkout from an SCM URL with optional tag
  > * Run the predefined Maven goals to release the project
  
  ([Source](http://maven.apache.org/maven-release/maven-release-plugin/examples/perform-release.html))

  In short: 
   * All shogun2 version-strings (in `pom.xml`) will be updated to the given version (e.g. `0.0.1`).
   * All modified files will be committed (and be included in a newly created tag, see below).
   * A new tag with the format `v{{NEW_VERSION}}` (e.g. `v0.0.1`) will be created (and will contain the actual state of the repo).
   * All shogun2 version-strings (in `pom.xml`) will be updated to the next development version (e.g.  `0.0.2`).
   * All modified files will be committed (and be included in the current master branch).
   * The newly created tag (or release) will be deployed to the staging area of the *Central Repository* repository.

  To do a release the following prerequisites are needed:
   * (maybe obsolute as we have a `shogun2` account) Create a Sonatype [JIRA account](https://issues.sonatype.org/secure/Signup!default.jspa)
   * Create a PGP signature and distribute your public key as described [here](http://central.sonatype.org/pages/working-with-pgp-signatures.html)
[...]

PR #132, opened 2016-02-17
Introduces the AbstractLayer and Layergroup

KaiVolland:
This introduces the AbstractLayer and Layergroup.

### DIFF
--- a/src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/layer/Layergroup.java
+++ b/src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/layer/Layergroup.java
@@ -1,7 +1,13 @@
 package de.terrestris.shogun2.model.layer;
 
+import java.util.ArrayList;
+import java.util.List;
+
 import javax.persistence.Entity;
-import javax.persistence.ManyToOne;
+import javax.persistence.JoinColumn;
+import javax.persistence.JoinTable;
+import javax.persistence.ManyToMany;
+import javax.persistence.OrderColumn;
 import javax.persistence.Table;
 
 import org.apache.commons.lang3.builder.EqualsBuilder;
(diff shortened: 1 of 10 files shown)


src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/layer/Layergroup.java
```diff
@@ -0,0 +1,134 @@
...
+ * and an appearance
+ *
+ * @author Kai Volland
+ * @author terrestris GmbH & Co. KG
+ *
+ */
+@Entity
+@Table
+public class Layergroup extends AbstractLayer {
+
```
buehner: CamelCase? -> `LayerGroup`

buehner:
I just had a very quick look at this, but as `AbstractLayer` is abstract, shouldn't the corresponding DAO/Service/Controller also be abstract (or are even not really necessary). I will have a deeper look at this tomorrow.

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/layer/Layergroup.java
KaiVolland: Isn't Layergroup one word?

KaiVolland:
I need a restinterface where i receive all layers and layergroups which are subclasses of AbstractLayer. I didn't test if i can realise that with abstract service and dao... I'll try that tomorrow.

chrismayer:
Hi @KaiVolland and @buehner,
do you think it will be possible to tag the current master as ``v0.0.4`` before merging this? What do you think? TIA!

KaiVolland:
@chrismayer Good idea. We changed a lot of things since the last version tag. Just do it.

buehner:
> Just do it.

I think that is easier said than done. As far as i know, currently only @dnlkoch und me can release in a way that assures that also the maven repos are updated (by using the release script, which requires some additional configuration like registering a SSH key on github).

But of course we can release `v0.0.4`. I will try to do this within the next hour. We should also document how such a release can be performed.

chrismayer:
Cool, thank you guys!

buehner:
You'll find some background information here: #113

buehner:
The `v0.0.4` release is finished (on the side of github). I think this version will appear at http://mvnrepository.com/artifact/de.terrestris/shogun2/0.0.4 within the next few hours

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/layer/Layergroup.java
buehner: Yes. In the end this is a matter of taste.

I prefer the CamelCase version as it is better readable. We should also handle this consistently within the project and as we already have a model `UserGroup`, i would also like to have `LayerGroup` here :wink:

chrismayer:
Hi @buehner,
many thanks for the release! I just checked mvnrepository.com and it seems that several artifacts got updated (e.g. [shogun2-service](http://mvnrepository.com/artifact/de.terrestris/shogun2-service)) and some not (e.g. [shogun2](http://mvnrepository.com/artifact/de.terrestris/shogun2/0.0.4)).

Just a matter of time or went something wrong?

buehner:
Hi @chrismayer 

this is really weird. It seems that this problem only occurs on http://mvnrepository.com/

On http://search.maven.org/ everything is released and seems to be fine.

We could not find any abnormalities in the logs, so i think we should wait some more time.

KaiVolland:
@chrismayer @buehner @weskamm Are there any further issues or can i merge this?

marcjansen:
Mvnreposititory.com now has 0.0.4 e.g. for the shogun package. 

http://mvnrepository.com/artifact/de.terrestris/shogun2

We are stil pre-2.0.0-final, and we have a clean way of getting the old behaviour by switching to e.g. 0.0.4. Sooooo:

Please merge.

PR #133, opened 2016-02-19
Get rid of cascading operations

buehner:
Until now, we had a vast quantity of `@Cascade(CascadeType.SAVE_UPDATE)` annotations in our models. This may not only lead to performance problems, but also resulted in serious problems when we tried to implement the removal of object references (where a deleted entity would have been recreated immediately due to an `@Cascade` annotation).

This PR removes ALL the `@Cascade` annotations and additionally simplifies the `ContentInitializer` a lot, which allows us to implement the above mentioned functionality.

The main difference of the "new" approach is that the `ContentInitializer` is now based on the following "generic" list of arbitrary `PersistentObject`s:

```Java
@Resource
@Qualifier("objectsToCreate")
private List<PersistentObject> objectsToCreate;
```

In this list (which is configured as a list of beans in the init-beans-XML from the archetype), the order of the elements is important as some objects may require the existence of related objects to avoid errors like `'object references an unsaved transient instance save the transient instance before flushing'`

Nobody wants to "forbid" the use of `@Cascade` annotations, but they should only be used where it really makes sense.

### DIFF
--- a/src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/SecuredPersistentObject.java
+++ b/src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/SecuredPersistentObject.java
@@ -9,8 +9,6 @@ import javax.persistence.OneToMany;
 import org.apache.commons.lang3.builder.EqualsBuilder;
 import org.apache.commons.lang3.builder.HashCodeBuilder;
 import org.apache.commons.lang3.builder.ToStringBuilder;
-import org.hibernate.annotations.Cascade;
-import org.hibernate.annotations.CascadeType;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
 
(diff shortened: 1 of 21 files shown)


KaiVolland:
Any idea what would happen if we had circular elements in the objectsToCreate-list? I think we should think about this now...

buehner:
I made a rebase on master and additionally removed all `@Cascade` annotations from the new classes. The archetype works fine.

Once travis is happy, this can be merged!

Be careful: Projects that are already using shogun2 need adaptions to work with the new bean/init implementation.

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/SecuredPersistentObject.java
```diff
@@ -29,12 +27,10 @@
 	private static final long serialVersionUID = 1L;
 
 	@OneToMany
```
KaiVolland: Please change this to `@ManyToMany` as we figured out that `@OneToMany` is wrong here.

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/SecuredPersistentObject.java
```diff
@@ -29,12 +27,10 @@
...
-	@Cascade(CascadeType.ALL)
 	@JsonIgnore
 	private Map<User, PermissionCollection> userPermissions = new HashMap<User, PermissionCollection>();
 
 	@OneToMany
```
KaiVolland: Please change this to `@ManyToMany` as we figured out that `@OneToMany` is wrong here.

KaiVolland:
@buehner I opened an PR against your branch for my comments, as i allready changed the annotations in my branch.

PR #141, opened 2016-03-02
HTTP helper utility

dnlkoch:
Adds simple HTTP utility methods for `GET` and `POST`.

### DIFF
--- a/src/shogun2-util/src/main/java/de/terrestris/shogun2/util/http/HTTPUtil.java
+++ b/src/shogun2-util/src/main/java/de/terrestris/shogun2/util/http/HTTPUtil.java
@@ -1,0 +1,464 @@
+package de.terrestris.shogun2.util.http;
+
+import java.io.IOException;
+import java.io.UnsupportedEncodingException;
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
+import java.net.Proxy;
+import java.net.ProxySelector;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.UnknownHostException;
+import java.util.Iterator;
+import java.util.List;
+
+import org.apache.http.Header;
+import org.apache.http.HttpEntity;
+import org.apache.http.HttpHost;
+import org.apache.http.NameValuePair;
+import org.apache.http.auth.AuthScope;
+import org.apache.http.auth.UsernamePasswordCredentials;
+import org.apache.http.client.AuthCache;
+import org.apache.http.client.CredentialsProvider;
+import org.apache.http.client.config.RequestConfig;
+import org.apache.http.client.entity.UrlEncodedFormEntity;
+import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.client.methods.HttpPost;
+import org.apache.http.client.methods.HttpRequestBase;
+import org.apache.http.client.protocol.HttpClientContext;
+import org.apache.http.entity.ContentType;
+import org.apache.http.entity.StringEntity;
+import org.apache.http.impl.auth.BasicScheme;
+import org.apache.http.impl.client.BasicAuthCache;
+import org.apache.http.impl.client.BasicCredentialsProvider;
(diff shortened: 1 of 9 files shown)


src/shogun2-util/src/main/java/de/terrestris/shogun2/util/http/HTTPUtil.java
```diff
@@ -0,0 +1,464 @@
...
+
+import de.terrestris.shogun2.model.Response;
+
+/**
+ *
+ * @author danielkoch
+ *
+ */
+public class HTTPUtil {
+
```
buehner: I would prefer `HttpUtil` (instead of `HTTPUtil`) :wink:

src/shogun2-util/src/test/java/de/terrestris/shogun2/util/http/HTTPUtilTest.java
```diff
@@ -0,0 +1,227 @@
...
+import de.terrestris.shogun2.model.Response;
+
+/**
+ *
+ * @author danielkoch
+ *
+ */
+@SuppressWarnings("static-method")
+public class HTTPUtilTest {
+
```
buehner: `HttpUtilTest` instead of `HTTPUtilTest` !? :wink:

dnlkoch:
Thanks for your review @buehner!

I addressed your notes and moved the plain non-persisted model object `Response` to a new util package `model` additionally. I suggest to use this package for any upcoming internal models.

src/shogun2-util/src/main/java/de/terrestris/shogun2/util/http/HttpUtil.java
```diff
@@ -0,0 +1,464 @@
...
+			httpRequest.setConfig(requestConfig);
+
+			httpResponse = httpClient.execute(httpRequest, httpContext);
+			HttpStatus httpStatus = HttpStatus.valueOf(
+					httpResponse.getStatusLine().getStatusCode());
+			String statusText = httpResponse.getStatusLine().getReasonPhrase();
+
+			if (httpStatus == HttpStatus.OK) {
+
+				response.setStatusCode(httpStatus);
```
marcjansen: I am not sure why we have this line.

src/shogun2-util/src/main/java/de/terrestris/shogun2/util/http/HttpUtil.java
```diff
@@ -0,0 +1,464 @@
...
+		} finally {
+
+			// cleanup
+
+			httpRequest.releaseConnection();
+
+			try {
+				if (httpResponse != null) {
+					httpResponse.close();
+				}
```
marcjansen: Can we use closeQuietly?

src/shogun2-util/src/main/java/de/terrestris/shogun2/util/http/HttpUtil.java
```diff
@@ -0,0 +1,464 @@
...
+	 * an InetSocketAddress ready to be used to pass the proxy to the
+	 * DefaultHttpClient.
+	 *
+	 * @param url
+	 * @return
+	 * @throws UnknownHostException
+	 */
+	private static HttpHost getSystemProxy(URI uri) throws UnknownHostException {
+
+		System.setProperty("java.net.useSystemProxies", "true");
```
marcjansen: I have always been wondering if we can set this only once, not for every call.

src/shogun2-util/src/main/java/de/terrestris/shogun2/util/http/HttpUtil.java
marcjansen: Ahh I see nowwe have our own wrapper class. Ignore the above note.

marcjansen:
Really great work, @dnlkoch! I have added som minor questions. Which you may want to look at.

Three more questions: 

* will https urls be fully supported? The code to determine a system wide proxy looks as if it is hardcoded for http.
* where will we handle allowed hosts? In the application?
* does this support other HTTP verbs besides POST and GET?

Again: a very well put together PR and the code looks really neat!

PR #142, opened 2016-03-03
Added groups to the user model / service

weskamm:
This PR adds usergroups to the usermodel. This way we can retrieve the assigned groups directly from the user instead of using the removed `findGroupsOfUser` function in the usergroup service

### DIFF
--- a/src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/User.java
+++ b/src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/User.java
@@ -8,16 +8,20 @@ import javax.persistence.Entity;
 import javax.persistence.JoinColumn;
 import javax.persistence.JoinTable;
 import javax.persistence.ManyToMany;
+import javax.persistence.OrderColumn;
 import javax.persistence.Table;
 
 import org.apache.commons.lang3.builder.EqualsBuilder;
 import org.apache.commons.lang3.builder.HashCodeBuilder;
 import org.apache.commons.lang3.builder.ToStringBuilder;
 
-import ch.rasc.extclassgenerator.Model;
-
+import com.fasterxml.jackson.annotation.JsonIdentityInfo;
+import com.fasterxml.jackson.annotation.JsonIdentityReference;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.ObjectIdGenerators;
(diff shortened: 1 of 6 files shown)


src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/User.java
```diff
@@ -8,17 +8,21 @@
...
 
-import ch.rasc.extclassgenerator.Model;
-
+import com.fasterxml.jackson.annotation.JsonIdentityInfo;
+import com.fasterxml.jackson.annotation.JsonIdentityReference;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.ObjectIdGenerators;
 import com.fasterxml.jackson.annotation.JsonProperty.Access;
 
+import ch.rasc.extclassgenerator.Model;
```
marcjansen: Is this needed?

marcjansen:
So users have groups and groups have users. Will they end up looking at one table? Or will the information be stored in different locations (which wouldn't be good, obviously).

Otherwise this looks fine.

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/User.java
```diff
@@ -51,6 +55,20 @@
 	)
 	private Set<Role> roles = new HashSet<Role>();
 
+	@ManyToMany
+	@JoinTable(
+		name = "user_usergroup",
```
KaiVolland: Givinig the JoinTable the same name leads to just one table. So the information is shared and not doubled.

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/UserGroup.java
```diff
@@ -37,6 +37,7 @@
 
 	@ManyToMany
 	@JoinTable(
+		name = "user_usergroup",
```
KaiVolland: Givinig the JoinTable the same name leads to just one table. So the information is shared and not doubled.

KaiVolland:
They share a table by giving the jointable the same name. See comments above.
`joinColumns` and `invertJoinColumns` have to be the opposite in the two classes.

marcjansen:
Very nice.

buehner:
LGTM. Feel free to merge

PR #149, opened 2016-03-17
Json identity info deserialization

buehner:
We are currently using the `@JsonIdentityInfo` annotation to serialize only the IDs of related objects (e.g. the groups of a user will be serialized by Jackson as an array of IDs, but not as complete objects.)

This PR handles the deserialization, i.e. the other direction by implementing
```Java
public class UserGroupIdResolver extends SimpleObjectIdResolver
```
and using this implementation in the `resolver` property of the `@JsonIdentityInfo` annotation.

Example: If a new User should be created, it is sufficient to POST an array of (existing) group IDs the user belongs to (in the `userGroups` property of the user object).

**Important note:** The current implementation only sets the ID value on a "new" UserGroup dummy object, which is enough information for Hibernate to persist the correct relation (in the join table). If we want access to full group objects, additional implementations are necessary (i suggest a follow up PR). E.g. the "load" method of the hibernate API could be a good choice here.

### DIFF
--- a/src/shogun2-core/shogun2-dao/pom.xml
+++ b/src/shogun2-core/shogun2-dao/pom.xml
@@ -23,6 +23,12 @@
             <version>${project.version}</version>
         </dependency>
 
+        <!-- Hibernate -->
+        <dependency>
+            <groupId>org.hibernate</groupId>
+            <artifactId>hibernate-core</artifactId>
+        </dependency>
+
         <!-- JUnit -->
         <dependency>
             <groupId>junit</groupId>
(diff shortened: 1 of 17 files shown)


buehner:
I added two commits that add missing ID resolvers for `Map` and `Layer`

buehner:
I added an ID resolver for `Layer` in the last commit

marcjansen:
Is this ready for review?

buehner:
Yes

dnlkoch:
LGTM!

KaiVolland:
Cool stuff. LGTM.

weskamm:
:+1:

KaiVolland:
@buehner plz merge

PR #150, opened 2016-03-17
Cascade nested (Layer-)models while saveOrUpdate

dnlkoch:
This PR suggests to (re-)add the `@Cascade(CascadeType.SAVE_UPDATE)` to all `Layer` associated models. In real world applications this seems to be very handy while saving a `Layer` (otherwise all child models would have to be saved manually first).


KaiVolland:
Thank you!

PR #151, opened 2016-03-24
Added Group REST Controller; extended User REST Controller and Service

weskamm:
This PR adds a REST Controller for UserGroups. The User and UserGroup REST Controller have been extended to support queries of subentities by pathvariables. This means you can now get the groups of a user by specifying the request like `rest/users/{id}/userGroups` or `rest/groups/{id}/users`

### DIFF
--- a/src/shogun2-web/src/main/java/de/terrestris/shogun2/rest/GroupRestController.java
+++ b/src/shogun2-web/src/main/java/de/terrestris/shogun2/rest/GroupRestController.java
@@ -11,35 +11,35 @@ import org.springframework.web.bind.annotation.RequestMethod;
 import org.springframework.web.bind.annotation.ResponseBody;
 import org.springframework.web.bind.annotation.RestController;
 
-import de.terrestris.shogun2.dao.UserDao;
+import de.terrestris.shogun2.dao.UserGroupDao;
 import de.terrestris.shogun2.model.User;
 import de.terrestris.shogun2.model.UserGroup;
-import de.terrestris.shogun2.service.UserService;
+import de.terrestris.shogun2.service.UserGroupService;
 import de.terrestris.shogun2.util.data.ResultSet;
 
 /**
- * @author Kai Volland
+ * @author Johannes Weskamm
  * @author Nils Bühner
  *
  */
 @RestController
-@RequestMapping("/users")
-public class UserRestController<E extends User, D extends UserDao<E>, S extends UserService<E, D>>
+@RequestMapping("/groups")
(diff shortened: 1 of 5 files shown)


src/shogun2-web/src/main/java/de/terrestris/shogun2/rest/GroupRestController.java
```diff
@@ -0,0 +1,76 @@
...
+import de.terrestris.shogun2.util.data.ResultSet;
+
+/**
+ * @author Johannes Weskamm
+ * @author Nils Bühner
+ *
+ */
+@RestController
+@RequestMapping("/groups")
+public class GroupRestController<E extends UserGroup, D extends UserGroupDao<E>, S extends UserGroupService<E, D>>
```
buehner: This should be named `UserGroupRestController` as our Model/Dao/Service is also using `UserGroup`.

src/shogun2-web/src/main/java/de/terrestris/shogun2/rest/GroupRestController.java
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	/**
+	 * Get the users of a specific group.
+	 *
+	 * @param groupId
+	 * @return
+	 */
+	@RequestMapping(value = "/{groupId}/users", method = RequestMethod.GET)
+	public @ResponseBody Map<String, Object> findUsersOfGroup(@PathVariable Integer groupId) {
```
buehner: As this is a REST interface we should return the entities, but not `Map<String, Objects`, which is based on `ResultSet.success(...)`, which is an Ext-JS specific format, isn't it?

src/shogun2-web/src/main/java/de/terrestris/shogun2/rest/UserRestController.java
```diff
@@ -46,4 +54,23 @@ protected UserRestController(Class<E> entityClass) {
...
 	}
+
+	/**
+	 * Get the groups of a specific user.
+	 *
+	 * @param userId
+	 * @return
+	 */
+	@RequestMapping(value = "/{userId}/userGroups", method = RequestMethod.GET)
+	public @ResponseBody Map<String, Object> findGroupsOfUser(@PathVariable Integer userId) {
```
buehner: As this is a REST interface we should return the entities, but not Map<String, Objects, which is based on ResultSet.success(...), which is an Ext-JS specific format, isn't it?

weskamm:
Issues adressed, thanks for review

weskamm:
merging this one now...

PR #153, opened 2016-04-01
Introduce util methods for HTTP PUT requests

ahennr:
This PR introduces methods for HTTP PUT requests to `HttpUtil` and their corresponding tests.

### DIFF
--- a/src/shogun2-util/src/main/java/de/terrestris/shogun2/util/http/HttpUtil.java
+++ b/src/shogun2-util/src/main/java/de/terrestris/shogun2/util/http/HttpUtil.java
@@ -496,7 +496,7 @@ public class HttpUtil {
 	 * @throws URISyntaxException
 	 */
 	public static Response put(URI uri) throws URISyntaxException, HttpException {
-		return putBody(new HttpPut(uri), StringUtils.EMPTY, ContentType.APPLICATION_JSON, null, null);
+		return putBody(new HttpPut(uri), null, null, null, null);
 	}
 
 	/**
@@ -508,7 +508,7 @@ public class HttpUtil {
 	 * @throws URISyntaxException
 	 */
 	public static Response put(String uriString) throws URISyntaxException, HttpException {
-		return putBody(new HttpPut(uriString), StringUtils.EMPTY, ContentType.APPLICATION_JSON, null, null);
+		return putBody(new HttpPut(uriString), null, null, null, null);
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


src/shogun2-util/src/main/java/de/terrestris/shogun2/util/http/HttpUtil.java
```diff
@@ -486,6 +488,138 @@ private static Response postParams(HttpPost httpRequest,
...
 	/**
+	 * Perform HTTP PUT with empty body
+	 *
+	 * @param uri
+	 * @return
+	 * @throws HttpException
+	 * @throws URISyntaxException
+	 */
+	public static Response put(URI uri) throws URISyntaxException, HttpException {
+		return putBody(new HttpPut(uri), StringUtils.EMPTY, ContentType.APPLICATION_JSON, null, null);
```
dnlkoch: Why do we set `ContentType.APPLICATION_JSON` here (and in the ones below) without having a body content?

src/shogun2-util/src/main/java/de/terrestris/shogun2/util/http/HttpUtil.java
```diff
@@ -486,6 +488,138 @@ private static Response postParams(HttpPost httpRequest,
...
+	 * @param username
+	 * @param password
+	 * @return The HTTP response as Response object.
+	 * @throws URISyntaxException
+	 * @throws HttpException
+	 */
+	private static Response putBody(HttpPut httpRequest, String body, ContentType contentType, String username, String password) throws URISyntaxException, HttpException {
+		StringEntity stringEntity = new StringEntity(body, contentType);
+		stringEntity.setChunked(true);
+		httpRequest.setEntity(stringEntity);
```
dnlkoch: Maybe you can introduce a condition here that checks for a not-null ContentType.

dnlkoch:
Very cool and useful addition! Thanks!

src/shogun2-util/src/main/java/de/terrestris/shogun2/util/http/HttpUtil.java
ahennr: According to issue in method `putBody(...)` further down, will pass `null` here for `body` and `ContentType`.

src/shogun2-util/src/main/java/de/terrestris/shogun2/util/http/HttpUtil.java
ahennr: Here, a body only will be included in the `HttpPut` request if it isn't `null` and has a valid `ContentType`

dnlkoch:
Great! Feel free to merge!

ahennr:
Thx for review, I'll merge!

PR #154, opened 2016-04-05
Remove EXT direct

buehner:
This PR completely removes EXT direct (and the EXT class generator _experiment_) from SHOGun2 because:

* it is complex and bloats up the project
* it has never been used (only in a very basic proof of concept ~2 years ago)
* it confuses
* it misleads developers to extend _wrong_ classes
* we do not intend to use it in future
* it does not support the latest EXT versions

Please review and merge. I do not expect that any adaptions in existing projects are necessary.


marcjansen:
Thanks for this, @buehner, great and thorough work with a good rationale. I agree we should remove this stuff.

Could you please start a tiny document that lists the places which existing projects (generated from previous archetypes) need to touch and change so that everything works against the newest snapshots?

The changes to the archetype here may be a good starting point. 

I suggest to document these places in a `RELEASE-NOTES.md`.

Once we have this, I am plus one on merging this PR.

buehner:
Thx @marcjansen 
I added a `RELEASE-NOTES.md` that describes what adaptions are necessary in existing projects.

Feel free to merge

marcjansen:
Thanks. I'll leave the merging up to you :smile: :+1:

PR #155, opened 2016-04-06
Fix hibernate annotations

buehner:
This PR eliminates the following WARNING:
`HHH000139: Illegal use of @Table in a subclass of a SINGLE_TABLE hierarchy`

On top of that, the missing `@AssociationOverrides` annotation has been added to the `AbstractLayer` class to ensure that user- and group permissions can be used correctly for the `AbstractLayer` hierarchy.

Existing projects should (once) boot with the `hibernate.hbm2ddl.auto` property set to `UPDATE` or `CREATE`


marcjansen:
Thanks! Good to go if checks go green. Can you add the notice

> Existing projects should (once) boot with the hibernate.hbm2ddl.auto property set to UPDATE or CREATE

to a file in the repo (just as I suggested in #154)?

buehner:
@marcjansen Yes, i can. Maybe the easiest way to do this would be to merge #154 
Then i could rebase this PR and add the information to the file.

buehner:
I just rebased and added the info to the release notes. Once this is green, we should merge it.

weskamm:
:+1: will merge

PR #156, opened 2016-04-06
Treeconfig

KaiVolland:
This introduces the TreeConfig.

Basicaly it is a representation of any tree **plus** these associations: 
- A Leaf has a connected layer
- A Folder has a connected layerGroup (as it is in GeoExt3)


src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/tree/TreeFolder.java
```diff
@@ -0,0 +1,168 @@
...
+	 * 
+	 */
+	@ManyToMany
+	@JoinTable(
+		name = "TREEFOLDER_CHILDREN",
+		joinColumns = { @JoinColumn(name = "TREEFOLDER_ID") },
+		inverseJoinColumns = { @JoinColumn(name = "TREENODE_ID") }
+	)
+	@OrderColumn(name = "IDX")
+	@Cascade(CascadeType.SAVE_UPDATE)
```
buehner: Cascade in a tree is ***shit, isn't it!?

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/tree/TreeNode.java
```diff
@@ -0,0 +1,125 @@
...
+ * The TreeNode is an abstract class which is the granular part of an tree.
+ *
+ * @author Kai Volland
+ * @author Daniel Koch
+ *
+ */
+@Entity
+@Table
+@Inheritance(strategy = InheritanceType.SINGLE_TABLE)
+public abstract class TreeNode extends PersistentObject{
```
buehner: Does @Table make sense with an abstract class!?

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/tree/TreeLeaf.java
```diff
@@ -0,0 +1,109 @@
...
+	/**
+	 * 
+	 */
+	@ManyToOne
+	@JsonIdentityInfo(
+			generator=ObjectIdGenerators.PropertyGenerator.class,
+			property="id"
+	)
+	@JsonIdentityReference(alwaysAsId=true)
+	private AbstractLayer layer;
```
buehner: Why does the TreeLeaf has an AbstractLayer, but the TreeFolder a LayerGroup (which is also an AbstractLayer)? Can't we pull this up to the common parent class?

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/tree/TreeFolder.java
```diff
@@ -0,0 +1,168 @@
...
+	/**
+	 * 
+	 */
+	@ManyToOne
+	@JsonIdentityInfo(
+			generator=ObjectIdGenerators.PropertyGenerator.class,
+			property="id"
+	)
+	@JsonIdentityReference(alwaysAsId=true)
+	private LayerGroup layerGroup;
```
buehner: Why does the TreeLeaf has an AbstractLayer, but the TreeFolder a LayerGroup (which is also an AbstractLayer)? Can't we pull this up to the common parent class?

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/tree/TreeNode.java
KaiVolland: @dnlkoch and I liked the idea of having all nodes in one table. No matter if it is a folder or a leaf.

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/tree/TreeLeaf.java
KaiVolland: The leaf should have a real `de.terrestris.shogun2.model.layer.Layer` i think. This would reflect the behaviour of geoext3.

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/tree/TreeFolder.java
KaiVolland: See above.

src/shogun2-core/shogun2-model/src/main/java/de/terrestris/shogun2/model/tree/TreeFolder.java
KaiVolland: I can't imagine any better example for a cascading behaviour then a tree. All the childnodes must be saved before saving a folder or we get transient issues.
Why do you think it is ***shit? (btw: what hard word do you hide with those three stars? :relaxed:  )

src/shogun2-core/src/main/java/de/terrestris/shogun2/model/tree/TreeNode.java
```diff
@@ -0,0 +1,157 @@
...
+	 *      http://stackoverflow.com/questions/27581/overriding-equals
+	 *      -and-hashcode-in-java it is recommended only to use getter-methods
+	 *      when using ORM like Hibernate
+	 */
+	@Override
+	public int hashCode() {
+		// two randomly chosen prime numbers
+		return new HashCodeBuilder(2, 89)
+				.appendSuper(super.hashCode())
+				.append(getText())
```
marcjansen: Missing `getType()`

src/shogun2-core/src/main/java/de/terrestris/shogun2/model/tree/TreeFolder.java
```diff
@@ -0,0 +1,131 @@
...
+	 *      http://stackoverflow.com/questions/27581/overriding-equals
+	 *      -and-hashcode-in-java it is recommended only to use getter-methods
+	 *      when using ORM like Hibernate
+	 */
+	@Override
+	public int hashCode() {
+		// two randomly chosen prime numbers
+		return new HashCodeBuilder(7, 89)
+				.appendSuper(super.hashCode())
+				.append(isExpanded())
```
marcjansen: Missing `getLayerGroup`, also for `equals` & `toString`

marcjansen:
I like it, please merge once my minor comments are adressed. Does @buehner want to say something as well for this updated version?

buehner:
I have some questions:

- What will be the root node of a TreePanel?
- How would we build a tree with depth > 1? Example:

  * (root)
    * folder
        * subfolder
         * leaf
         * leaf
    * leaf

KaiVolland:
The concept of this PR is, that we just save some additional informations to already existing tree nodes:

- checked/unchecked
- collpased/expanded
- custom text

The tree itself is not created by this entities!

If you want to create a tree with a depth larger one you need to create a LayerGroup that contains a LayerGroup (...). You can then add this layergroup to the mapmodule.

buehner:
OK. I am just asking because i am (still) confused. When i saw "Tree", i was looking for something like _rootNode_, a list of _children_ or any kind of recursive structure.

So would you agree that some classes of this PR should be renamed to be more meaningful?
Or should they even be a project solution!?

I don't want to block the merge of this PR, but i am unsure about this issue - maybe because i don't understand the whole concept here.

Just to be sure that i understood everything correctly:

  * (root)
    * Folder A
        * Folder B
         * Leaf A
         * Leaf B
    * Leaf C

Would a `TreePanel` for this scenario look like this:

  * `TreePanel`
    * nodeConfigs = [`TreeFolder` for Folder A; `TreeNode` for Leaf C]

buehner:
OK. I am just asking because i am (still) confused. When i saw "Tree", i was looking for something like _rootNode_, a list of _children_ or any kind of recursive structure.

I don't want to block your work, but i am unsure about this issue - maybe because i don't understand the whole concept here.

Just to be sure that i understood everything correctly:

  * (root)
    * LayerGroup A
        * LayerGroup B
         * Layer A
         * Layer B
    * Layer C

Would a `TreePanel` for this scenario look like this:

  * `TreePanel`
    * nodeConfigs = [`TreeFolder` for LayerGroup A; `TreeNode` for Layer C]

where LayerGroup A would have LayerGroup B, Layer A and Layer B as children.

In such a scenario, we could not control whether LayerGroup B should initially be extended or not or whether Layer A or B should be checked or unchecked. We could only control this for the first level, i.e. for LayerGroup A and Layer C. Is this correct?

If yes, then this would be a major drawback of this design because we could only control the checked/expanded state of nodes on the first level and then this should be a project solution (if this satisfies the requirements of the project).

When i compare the `AbstractLayer`/`LayerGroup`/`Layer` with the proposals of this PR, i have the feeling that there is a mixture of concepts. It seems that the `AbstractLayer` hierarchy is already representing the tree structure that i would expect here (but not in the layer hierarchy). I think we should take some time in the future to review and overhaul the current status. Maybe we can get rid of the `LayerGroup` model and should only provide a simple `Layer` model at this point. Then we could model a recursive `TreeNode` hierarchy that makes use of these simple `Layer` representations.

KaiVolland:
As the shogun2-core-devs are not 100% satisfied with this i move it to my project code.

PR #157, opened 2016-04-07
Simplify project structure

buehner:
This PR simplifies the whole project structure by merging all modules into the shogun2-core module.

The main reason for this step is to avoid cyclic dependencies - a scenario we had at some points in the past weeks. On top of that, the development will be simplified as there is only one project left in the IDE.

In future, we should think about senseful modules we possibly want to extract, e.g. shogun2-interceptor, shogun2-importer, shogun2-rest...

This PR also contains an update to the latest versions of the used dependencies.

Feel free to merge.


KaiVolland:
If you think this is the best way I'm ok with it.

KaiVolland:
hoppsa

dnlkoch:
Go for it!

marcjansen:
I like it. Next time I'd prefer to have separate PRs for the version updates and other changes.

PR #159, opened 2016-04-08
Load entities from the DB when resolving by IDs

buehner:
Here is a nice enhancement of the current implementation of the Property-/ID-Resolver classes:

We are using the `@JsonIdentityInfo` annotation to serialize objects as their ID representation. For the other direction, we already have a basic implementation of ID resolvers to deserialize such IDs to POJOs. But the current implementation only creates some "dummy" objects that only contain the ID value (but nothing else). This is sufficient to assign objects to Collections (as only the IDs are persisted in some join tables), but the current status may lead to problems with HashCodes/Sets and one would have to "manually" load such objects from the database to get full access to the object.

This PR enhances the existing ID resolvers by loading the correct entity from the database, which means that you can send IDs to the backend, but in the first line of your code, you will have the full entity available!

Feel free to merge.

### DIFF
--- a/src/shogun2-core/src/main/java/de/terrestris/shogun2/converter/LayerIdResolver.java
+++ b/src/shogun2-core/src/main/java/de/terrestris/shogun2/converter/LayerIdResolver.java
@@ -1,30 +1,25 @@
 package de.terrestris.shogun2.converter;
 
+import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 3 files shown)


buehner:
Could anyone review please? @marcjansen @dnlkoch @KaiVolland @weskamm

marcjansen:
Looks good to me. Please merge and thanks.

weskamm:
really nice addition :+1:

PR #164, opened 2016-04-18
Introduce AbstractSecuredPersistentObjectService

buehner:
This PR introduces an AbstractSecuredPersistentObjectService that provides four methods for an easy handling of entity permissions:

* `addUserPermissions(E entity, User user, Permission... permissions)`
* `removeUserPermissions(E entity, User user, Permission... permissions)`
* `addGroupPermissions(E entity, UserGroup userGroup, Permission... permissions)`
* `removeGroupPermissions(E entity, UserGroup userGroup, Permission... permissions)` 

These methods can easily be called to add or remove permissions for an entity. They will use an existing permission collection or create new ones if necessary. They will also persist the permission information.

All services of entities of type `SecuredPersistentObject` should extend this new abstract service

This commit also removes the PermissionsUtil which is obsolete due to the new service.

Please review and feel free to merge!

PR #165, opened 2016-04-19
Optimize permission handling

buehner:
Based on the comments of @marcjansen in #164, this PR optimizes the handling of permissions in the `AbstractSecuredPersistentObjectService`:

* Renamed methods to be more meaningful (e.g. `addAndSaveUserPermissions(...)`)
* Only persist permission collections, if the number of elements in the set has really changed

I also added tests for the new case.

Please review and merge

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,6 +1,10 @@
 # Release Notes
 
 ## 0.1.0 (not finally released yet)
+
+* New features:
+  * An `AbstractSecuredPersistentObjectService` has been introduced. This service provides useful methods to add and remove permissions for certain objects. `PermissionCollection`s will be persisted in the database when using these methods. All services of entities that extend `SecuredPersistentObject` should extend the abstract service mentioned above.
+
 Existing projects (that were possibly created with an old version of the webapp archetype) need adaptions regarding the following points:
 * Adapt the ``pom.xml`` of your existing SHOGun2 project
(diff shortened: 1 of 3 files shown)


marcjansen:
Thanks for taking on this thing again.

Just one purely stylistic thing:

In all the methods you correctly check that `entity` is not null, and only then do something inside the `if`, while in the `else` case you log an error:

```java
if (foo != null) {
    // do sth.
} else {
    // log that foo was null
}
```

I'd restructure this as follows (But mind me, it is a personal preference):

```java
if (foo == null) {
    // log that foo was null
    return;
}
// do sth.
```

That saves one level of indentation for the actual logic.

You can change it if you agree, but I have no strong feeling about the above, I just think it enhances readability a lot.

Please merge at will.

buehner:
@marcjansen You're absolutely right regarding the code style of the if/else blocks. I changed this in 026bc3f. Feel free to merge

weskamm:
returning early, saving indentation, :+1:

marcjansen:
Should we think about using e.g. https://commons.apache.org/proper/commons-lang/javadocs/api-3.4/org/apache/commons/lang3/Validate.html or assertions generally?

weskamm:
I think that discussion has to take place in the issues section, as it affects the whole project.
Are we ready to merge?

buehner:
Yes!

PR #167, opened 2016-04-21
Use full user object when evaluating permissions

buehner:
This PR is an enhancement of the current permission evaluation.

Until now, we were always evaluating permissions based on a given user ID. This PR changes the code to make use of the full user object (coming from the db).

To realize this, we have to use the `UserDao` in the `Shogun2PermissionEvaluator` class. We can NOT use the `UserService` here as the methods of the user service are secured, which would trigger the permission evaluation again and we would always end up with a StackOverflow error when using the service here.

These changes do not require any adaptions in existing projects!

Please review

### DIFF
--- a/src/shogun2-core/src/main/java/de/terrestris/shogun2/security/access/Shogun2PermissionEvaluator.java
+++ b/src/shogun2-core/src/main/java/de/terrestris/shogun2/security/access/Shogun2PermissionEvaluator.java
@@ -3,9 +3,12 @@ package de.terrestris.shogun2.security.access;
 import java.io.Serializable;
 
 import org.apache.log4j.Logger;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.security.access.PermissionEvaluator;
 import org.springframework.security.core.Authentication;
 
+import de.terrestris.shogun2.dao.UserDao;
 import de.terrestris.shogun2.model.PersistentObject;
 import de.terrestris.shogun2.model.User;
 import de.terrestris.shogun2.model.security.Permission;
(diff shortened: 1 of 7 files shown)


src/shogun2-core/src/main/java/de/terrestris/shogun2/security/access/Shogun2PermissionEvaluator.java
```diff
@@ -23,6 +26,18 @@
 	 */
 	private final static Logger LOG = Logger.getLogger(Shogun2PermissionEvaluator.class);
 
+	/**
+	 * We have to use the DAO here. If we would use the service, we would end
+	 * with StackOverflow errors as a call to (secured) service methods triggers
+	 * this PermissionEvaluator class.
+	 */
```
marcjansen: Nice find!

src/shogun2-core/src/main/java/de/terrestris/shogun2/security/access/Shogun2PermissionEvaluator.java
```diff
@@ -42,7 +57,12 @@ public boolean hasPermission(Authentication authentication,
 				&& targetDomainObject instanceof PersistentObject
 				&& permissionObject instanceof String) {
 
-			final User user = (User) authentication.getPrincipal();
+			// get the user state when the user logged in
+			User user = (User) authentication.getPrincipal();
```
marcjansen: Can this ever be `null`?

marcjansen:
Very nice simpolification. I added one question, which you may want to answer.

Thanks and please merge at will, @buehner.

src/shogun2-core/src/main/java/de/terrestris/shogun2/security/access/Shogun2PermissionEvaluator.java
buehner: I don't think so, but this should not be problem here as we check for `authentication.getPrincipal() instanceof User` a few lines before

marcjansen: :+1:

PR #168, opened 2016-04-22
Fix Layer hierarchy

buehner:
The `AbstractLayer` hierarchy had some drawbacks that should be fixed with this PR.

* General bugfixing (e.g. in constructors)
* The layer services now all extend from `AbstractSecuredPersistentObjectService`, which allows us to make use of the very useful methods from this class
* The methods from the `AbstractLayerService/DAO`  have been moved to the correct (and new) subclass `LayerGroupService/DAO`
* Missing services and controllers and DAOs have been added

Please review


weskamm:
Looking good to me, of course we should still discuss the TODOs and Exception warning internally

PR #170, opened 2016-04-25
Fix hasPermission check on findMapsWithLayer method

weskamm:
Usage of `#e` was wrong here, as the map module does not extend from SecuredPersistentObject.
The layer does, and thats what we check now!


buehner:
nice catch. thx
i'll merge now

PR #171, opened 2016-04-25
Use findById instead of loadById when deleting

buehner:
The `loadById` method creates proxy objects which lead to problems when using the REST-Delete method. Using `findById` fixes this.

Please review.


marcjansen:
👍

buehner:
Just to clarify: REST/delete is working fine in the current state. But if you overwrite the delete method in a specific scenario this may be problematic, e.g. if you try to remove the object (which is only a proxy in case of `loadById`) from an existing Set.

weskamm:
:+1: , merge at will

PR #174, opened 2016-04-26
Prepare interceptor beans in archetype resources

marcjansen:
This makes it easier for people to actually use the already existing package
classes for the various interceptors; they simply have to uncomment some lines
and are then ready to implement project or artifact specific interceptor
behaviour.

Please review.

/cc @buehner @dnlkoch


dnlkoch:
Nice addition! Thx!

marcjansen:
Cheerio for the review!

PR #176, opened 2016-04-26
Make saveOrUpdate void

buehner:
* The hibernate `saveOrUpdate` method is void.
* Our DAO `saveOrUpdate` method is void.
* Our service `saveOrUpdate` is **NOT** void, but returns the persisted object, which is not necessary and not consistent with the two methods mentioned above. On top of that it misleads developers to create new objects unnecessarily - even though the existing entity (on which `saveOrUpdate` is called) can easily be used in next steps.

This PR makes the `saveOrUpdate` of the services void.

Existing projects may need simple adaptions like changes

from `PersistentObject newObject = oldObject.saveOrUpdate()`
to `oldObject.saveOrUpdate()`

Please review

### DIFF
--- a/src/shogun2-core/src/test/java/de/terrestris/shogun2/rest/AbstractRestControllerTest.java
+++ b/src/shogun2-core/src/test/java/de/terrestris/shogun2/rest/AbstractRestControllerTest.java
@@ -9,6 +9,7 @@ import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.verifyNoMoreInteractions;
 import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.doAnswer;
 import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.delete;
 import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
 import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.post;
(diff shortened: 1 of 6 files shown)


src/shogun2-core/src/test/java/de/terrestris/shogun2/rest/AbstractRestControllerTest.java
```diff
@@ -221,13 +224,19 @@ public void findById_shouldReturn_NotFound() throws Exception {
...
 		TestModel withoutId = buildTestInstanceWithValue(value);
-		TestModel withId = buildTestInstanceWithIdAndValue(id, value);
 
-		when(serviceMock.saveOrUpdate(any(TestModel.class))).thenReturn(withId);
+		doAnswer(new Answer<Void>() {
+			@Override
+			public Void answer(InvocationOnMock invocation) throws Throwable {
+				TestModel entity = (TestModel) invocation.getArguments()[0];
+				IdHelper.setIdOnPersistentObject(entity, id);
+				return null;
```
marcjansen: Is that the way how you fulfill the expected return type `void`? Can you change this to

```java
return;
```

or remove the `return` line entirely?

This would make this easier to read IMO.

marcjansen:
Looks good to me  @buehner, I like consistency. The comment wrt to `return` in the tests applies to all `Answer` mocks.

I'll let you decide what to do about the comment, please merge at will with or without changing the code.

weskamm:
also :+1: here

src/shogun2-core/src/test/java/de/terrestris/shogun2/rest/AbstractRestControllerTest.java
buehner: Neither the use of `return;` nor the removal of the whole line will work here.

So this seems to be the way as described here: http://stackoverflow.com/a/2278221

PR #179, opened 2016-04-26
Changes registration methods from private/protected to public

KaiVolland:
I need to access theses methods from outside. Alternatives welcome.


marcjansen:
Commit history doesn't look well. We can sort it out together.

KaiVolland:
I'm more interested in feedback to the code. But if this is ok we can cleanup history...

chrismayer:
Hi @KaiVolland, do you have more information waht are the reasons to change the visibilitiy to public?

KaiVolland:
It's projectspecific stuff where i have to create users from outside of the userservice.
Therefore i need to get a token and send a registration mail.

chrismayer:
OK, then you can change visibility as you did or you can introduce public wrapper methods.

buehner:
Both methods are called from the `sendRegistrationActivationMail` in the `RegistrationTokenService`.

If you want to send a registration activation mail. Wouldn't that be a method you could use?

buehner:
Maybe it makes sense to refactor the `sendRegistrationActivationMail` method to accept a mail message (instead of always using the `registrationMailMessageTemplate`)!?

KaiVolland:
No cause this method sends an email itself. I need just the url and the token to put it in my projectspecific mail.

Are there any security-concerns etc. with my suggested changes?
If yes, a refactoring of the `RegistrationTokenService` would be ok for me.
If no, i would just like to see this merged.

buehner:
If you want a projectspecific mail content, then please change the `sendRegistrationActivationMail` method to accept a `SimpleMailMessage` object with your specific mail content. The current implementation always uses the injected `registrationMailMessageTemplate`.

I am not sure if we really want to make the `getValidTokenForUser` method public. For example, the `buildConcreteInstance` method is called there (which is an abstract method without implementation) that has to be implemented by subclasses (like the `RegistrationTokenService` does).

KaiVolland:
I refactored the Service. So just `createRegistrationActivationURI` is public now. It now receives an user instead of a token.

buehner:
:+1: merge if you want

KaiVolland:
Thanks

marcjansen:
Nice discussion here, everyone 👏👏👏

PR #180, opened 2016-04-28
Allow creation and updating of permissions

buehner:
The `PermissionCollection` type is unsecured, but the `PersistentObjectPermissionEvaluator` only grants `READ` access on unsecured objects by default. (Otherwise one could misuse some REST-interfaces to create unsecured objects, which is not desired).

In case of `PermissionCollection` it is necessary to allow that they can be created and updated. For that reason the new class `PermissionCollectionPermissionEvaluator` has been introduced (where `hasPermission` always returns true).

Whether or not a `PermissionCollection` may be created or updated depends on the entity for which the permission collection is used. For this, i suggest to secure the methods of `AbstractSecuredPersistentObjectService` in a follow up-PR


weskamm:
great stuff, please merge when green

PR #186, opened 2016-05-10
Protect by id and type of entity

buehner:
This PR introduces an implementation of the
```Java
public boolean hasPermission(Authentication authentication, Serializable targetId, String targetType, Object permission)
```
which is based on the ID and the type of a target entity.

The implementation provided here searches for a matching DAO, retrieves the entity from the database and simply calls the (existing) implementation of `hasPermission` with the object.

Two methods in the `MapService` and the `LayerGroupService` now make use of the new implementation.

Please review

### DIFF
--- a/src/shogun2-core/src/main/java/de/terrestris/shogun2/security/access/Shogun2PermissionEvaluator.java
+++ b/src/shogun2-core/src/main/java/de/terrestris/shogun2/security/access/Shogun2PermissionEvaluator.java
@@ -1,13 +1,16 @@
 package de.terrestris.shogun2.security.access;
 
 import java.io.Serializable;
+import java.util.Collection;
 
 import org.apache.log4j.Logger;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.context.ApplicationContext;
 import org.springframework.security.access.PermissionEvaluator;
 import org.springframework.security.core.Authentication;
 
+import de.terrestris.shogun2.dao.GenericHibernateDao;
 import de.terrestris.shogun2.dao.UserDao;
 import de.terrestris.shogun2.model.PersistentObject;
 import de.terrestris.shogun2.model.User;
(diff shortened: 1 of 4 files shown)


weskamm:
Although the title seems to be missing something, this PR adds a nice new flexibility!
I like it, :+1:

src/shogun2-core/src/main/java/de/terrestris/shogun2/security/access/Shogun2PermissionEvaluator.java
```diff
@@ -85,12 +91,76 @@ public boolean hasPermission(Authentication authentication,
...
+			if(dao.getEntityClass().equals(entityClass)) {
+				// we found a matching DAO
+				daoToUse = dao;
+				LOG.debug("Found an exactly matching DAO for type " + entityClass);
+				break;
+			}
+		}
+
+		// if we could not find an exact match, we'll try to use the "next best"
+		// from the entity hierarchy
```
marcjansen: This will possibly give you true for a class that isn't the best choice, right?

Consider:

```
A.java (With a matching DAO)
⇧
B.java (With a matching DAO)
⇧
C.java (With a matching DAO)
⇧
D.java (No DAO)
```

If entityClass is an instance of `D`, we'll drop here, as we do not have a direct dao-match. The best bet for a dao would be the `C-Dao`, but when we iterate, `isAssignableFrom` will be true for any of the parent classes (btw. their daos). We might end up with the not-well-suited A-Dao, for example.

Did I understand this correctly? Amnd if so, can `getSuperclass()` help? We could then factor ot the doa finding into a small methdi that we could call reapeatedly unto i we actually have the best bet. What do you think?

marcjansen:
Pleas give this PR a better title and also please take the time to think about the longer remark I added to the diff.

Regardless of whether you take the suggestion, this is good to go IMO. Thanks @buehner. Please merge at will (taking or ignoring my suggestion).

buehner:
Yes. You're right @marcjansen 

But this will probably never be a "real" problem as we only need a DAO to call `findById` (which will most likely always be the same implementation).

So i will merge this PR now and open a new issue to adress your proposal.

buehner:
I just opened #187

PR #188, opened 2016-05-11
Making the REST update method transactional

weskamm:
See the comment in the diff, which explains what and why we do this.
It makes sure that entities do not get persisted by mistake due to `FlushMode.AUTO` in hibernate.

### DIFF
--- a/src/shogun2-core/src/main/java/de/terrestris/shogun2/service/AbstractCrudService.java
+++ b/src/shogun2-core/src/main/java/de/terrestris/shogun2/service/AbstractCrudService.java
@@ -1,10 +1,15 @@
 package de.terrestris.shogun2.service;
 
+import java.io.IOException;
 import java.util.List;
 
 import org.springframework.security.access.prepost.PostAuthorize;
 import org.springframework.security.access.prepost.PostFilter;
 import org.springframework.security.access.prepost.PreAuthorize;
+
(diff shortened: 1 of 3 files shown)


buehner:
nice catch!

:+1: from me

i think this is the related documentation of hibernate (FlushMode.Auto -> autosync of managed entities that have been changed in the meantime): http://docs.jboss.org/hibernate/orm/current/userguide/html_single/Hibernate_User_Guide.html#flushing

src/shogun2-core/src/main/java/de/terrestris/shogun2/service/AbstractCrudService.java
```diff
@@ -40,9 +45,24 @@ public void saveOrUpdate(E e) {
 	}
 
 	/**
+	 * @param jsonObject
+	 * @param entity
+	 * @return
+	 * @throws IOException
+	 * @throws JsonProcessingException
+	 */
+	@PreAuthorize("hasRole(@configHolder.getSuperAdminRoleName()) or hasPermission(#e, 'UPDATE')")
```
buehner: Important: `#e` must be `#entity`

buehner:
Please fix the `#e` bug and merge if you are happy :wink:

marcjansen:
Approved!

So glad this is fixed! Thanks a ton.

weskamm:
Thanks guys, merging now!

PR #189, opened 2016-05-13
Adds missing ogc EndPoint "typename".

KaiVolland:
Adds missing ogc EndPoint "typename"


marcjansen:
Great catch, please merge once green and thanks @KaiVolland.

PR #190, opened 2016-05-13
Introduce methods to find permissions of a User or UserGroup

buehner:
These new methods are helpful to find out on which entities (of the whole system) a user or user group has permissions.

The current implementation is not optimal (see comments in code) and should be used with care.

Please review and merge


weskamm:
Great Helper stuff, i thank you my friend

PR #192, opened 2016-05-24
Adapt the permission evaluation to be less restrictive

buehner:
This PR adapts the permission evaluation to be less restrictive. This is necessary in scenarios where NO user is logged in (ROLE_ANONYMOUS), but we still want the permission evaluators to run.


KaiVolland:
LGTM.

PR #193, opened 2016-06-01
Introduces HTTP DELETE to HttpUtil.

KaiVolland:
Title says it all.

### DIFF
--- a/src/shogun2-core/src/main/java/de/terrestris/shogun2/util/http/HttpUtil.java
+++ b/src/shogun2-core/src/main/java/de/terrestris/shogun2/util/http/HttpUtil.java
@@ -85,6 +85,8 @@ public class HttpUtil {
 	 * Performs an HTTP GET on the given URL.
 	 *
 	 * @param url The URL to connect to.
+	 * @param username The Basic authentication username.
+	 * @param password The Basic authentication password.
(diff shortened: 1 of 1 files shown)


src/shogun2-core/src/main/java/de/terrestris/shogun2/util/http/HttpUtil.java
```diff
@@ -600,6 +601,67 @@ public static Response put(URI uri, String body, ContentType contentType) throws
...
+	 * Performs an HTTP DELETE on the given URL.
+	 *
+	 * @param url The URL to connect to.
+	 *
+	 * @return The HTTP response as Response object.
+	 *
+	 * @throws URISyntaxException
+	 * @throws HttpException
+	 */
+	public static Response delete(String url, String username, String password)
```
buehner: JavaDoc:

`@param username` and `@param password` is missing

:wink:

buehner:
I just mentioned a small issue regarding the JavaDoc. The `@param username` and `@param password` is also missing for the methods

* `public static Response get(String url, String username, String password)`
* `public static Response post(String url, String username, String password)`

Maybe you can also fix this there?

Feel free to merge afterwards. Thx

PR #195, opened 2016-06-02
Cleanup layer hierarchy

buehner:
The current hierarchy of `AbstractLayer` in the model package is very questionable because the `LayerGroup` subclass represents some kind of Layer tree (folder) concept, which should be modeled separately. (A PR for a `TreeNode` hierarchy will follow in the near future).

This PR simplifies this hierarchy by removing the `LayerGroup` concept. Instead, there will only be a simple `Layer` model (which may be extended in projects if necessary).

Please review.


src/shogun2-core/src/main/java/de/terrestris/shogun2/model/layer/Layer.java
```diff
@@ -49,17 +65,28 @@ public Layer() {
...
 	/**
 	 * @param name Layer name
-	 * @param type Layer type
-	 * @param source The data source of the layer
-	 * @param appearance The appearance configuration of the layer
 	 */
-	public Layer(String name, String type, LayerDataSource source, LayerAppearance appearance) {
-		super(name, type);
-		this.source = source;
-		this.appearance = appearance;
```
dnlkoch: Minor: Why do you removed the constructor arguments `source` and `appearance`?

dnlkoch:
Great job! :+1:

src/shogun2-core/src/main/java/de/terrestris/shogun2/model/layer/Layer.java
buehner: I think i just missed that during the refactoring process. I added this constructor again...

src/shogun2-webapp-archetype/src/main/resources/archetype-resources/src/main/resources/META-INF/spring/__artifactId__-context-initialize-beans.xml
```diff
@@ -425,14 +425,12 @@
 
     <bean id="osmWmsLayer" class="de.terrestris.shogun2.model.layer.Layer">
         <property name="name" value="OSM-WMS" />
-        <property name="type" value="Tile" />
```
KaiVolland: Are you shure that this is not used anywhere?

KaiVolland:
One question regarding `type` in Layer.

src/shogun2-webapp-archetype/src/main/resources/archetype-resources/src/main/resources/META-INF/spring/__artifactId__-context-initialize-beans.xml
buehner: Yes. Because there is no `type` property in the `Layer` model in SHOGun2

marcjansen:
Note to ourselves: Can we try to release a version before such big changes?

marcjansen:
Another note: I want to discuss whether we may need tests for our modelling of these concepts. I know this isn't common, but I fear it would force us to be more restrictive with modelling and changes to it.

The change itself is much appreciated, don't get me wrong.

src/shogun2-webapp-archetype/src/main/resources/archetype-resources/src/main/resources/META-INF/spring/__artifactId__-context-initialize-beans.xml
chrismayer: This removement raises an issue in SHOGun2-client project while [creating layer sources](https://github.com/terrestris/shogun2-client/blob/master/app/view/component/MapController.js#L133). I know SHOGun2-client is a separate project but somehow related.

marcjansen: Thanks for mentioning this. Can you create an issue there?

chrismayer: Done, see https://github.com/terrestris/shogun2-client/issues/16

PR #196, opened 2016-06-02
Configurable HTTP connection timeout

dnlkoch:
This PR introduces a configurable HTTP connection timeout.


buehner:
Could you test this successfully? If yes: :+1: and merge please.

Existing projects then need to introduce the `http.timeout` property.

marcjansen:
> Existing projects then need to introduce the http.timeout property.

Aternatively, can we fall-back to the previous value in case `${http.timeout}` evaluates to a 'falsy' / 'undefined' value, maybe inside of `setHttpTimeout`?

dnlkoch:
Yes, this is potentially possible, for example we could easily make use of a default/fallback value with a simple expression in the `@Value` annotation itself: `@Value("${http.timeout:30000}")`.

marcjansen:
Yeah! That sounds very cool!

PR #198, opened 2016-06-13
Introduce a territory class

buehner:
This PR introduces a new model class `Territory` that contains a geometry property of type `MultiPolygon` (JTS). This new class may be very helpful in project solutions, for example if you want to store information about spatial restrictions for a certain user or user group.

Currently, the geometry property will be saved as binary data in the database (Java serialized representation), which is not a very smart, but working solution. For the future, we could think about using Hibernate Spatial to persist the geometry, but this would require that the underlying SHOGun2 database is always a spatial database.

As we are using the Java Topology Suite, this PR also introduces a new dependency on `jackson-datatype-jts`, which provides a `JtsModule` that is now registered in the `Shogun2JsonObjectMapper`. By doing this, JTS-geometries will always be serialized as GeoJSON (and vice versa).

### DIFF
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -184,6 +184,12 @@
         <!-- Powermock -->
         <powermock.version>1.6.4</powermock.version>
 
+        <!-- Java Topology Suite -->
+        <jts.version>1.14.0</jts.version>
+
+        <!-- JTS Jackson Datatype -->
+        <jackson-datatype-jts.version>2.3</jackson-datatype-jts.version>
+
         <downloadSources>true</downloadSources>
         <downloadJavadocs>true</downloadJavadocs>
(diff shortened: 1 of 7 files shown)


marcjansen:
So currently it would be impossible to actually visualize the territories, right? Or can we e.g. point a GeoServer towards the generated table and have it visualize the territories?

Can we please at least add a TODO and probably also an associated github issue? Also a detailed example (in the archetype?) would be cool. If we cannot have this, an issue and/or TODO is much appreciated.

Otherwise this looks good to me, please merge.

buehner:
I opened a corresponding issue and will merge this PR now: https://github.com/terrestris/shogun2/issues/199

PR #200, opened 2016-06-16
Adapt the logging of the interceptor

buehner:
The interceptor is logging a lot on the DEBUG level. I think some information could only be logged on the TRACE level.


marcjansen:
Agreed. Please merge

PR #201, opened 2016-06-17
Use request in response interceptor

buehner:
All Response-interceptors from the package `de.terrestris.shogun2.util.interceptor` now have a new parameter `MutableHttpServletRequest request`. This makes sense as the request object contains basic information that may be necessary/useful in custom implementations of a response interceptor.

There is also a small change in the static util methods of `MutableHttpServletRequest`, so that these methods can also be used with a `HttpServletRequest`.

This PR also adapts the tests and the release notes. Please review.


marcjansen:
Can you fix the build? I am +1 once the checks are green

KaiVolland:
What about this PR?

marcjansen:
Good to go IMO

PR #202, opened 2016-06-27
Fix interceptor bugs

buehner:
This is an enhancement for the interceptor:

* There was a bug when a HTTP (GET) request had multiple, comma-separated values in one parameter, e.g. `LAYERS=A,B`, which resulted in `LAYERS=AB`. de3c922 fixes this.
* 6e27bda is a more solid way to check for the existance of a key. the previous version is not save as a map could also contain `null` as a value...

Please review


marcjansen:
Cool, thanks. I would have prefered two PRs and possibly tests but I am nonetheless `+1`

PR #203, opened 2016-07-05
Fixed the password reset workflow

weskamm:
This PR fixes the currently broken password reset workflow.

1.) The `PasswordResetTokenService` now uses the `userDao` in method `sendResetPasswordMail` as the user who triggers this request cannot login because of his lost password and therefore has no authentication, which would be needed when using the service methods.

2.) The `@PreAuthorize` annotation in method `updatePassword` of `UserService` has been removed completely, as this method is also triggered by a non-logged-in user with no authentication.

Finally, minor typos have been fixed.

I am open to discussion if this is the way to go, but some projects are currenlty broken by the current state


buehner:
Using the `userDao` in this scenario (without authentication) makes sense!

Regarding the `updatePassword` method: This seems a bit dangerous, but as long as this method is only called by a controller that accepts a token (which _secures_ the process here), it should be OK.

:+1:

weskamm:
Thanks, merging

PR #204, opened 2016-07-08
Removed transfer-encoding:chunked header in httputil

weskamm:
The HttpClient automatically handles the chunked transfer-encoding when requesting resources, and returns non chunked content. But currently we copy all headers from the original response to the final response, which declarates the answer as chunked, which it is not at that moment. So we do not set the header for responses, if present in original response


buehner:
nice catch. :+1: 

feel free to merge, if travis is happy.

weskamm:
tahnks

dnlkoch:
+1!

Issue #205, opened 2016-07-14
Refactoring Trees


KaiVolland:
As we allready decided we will redesign or introduce a general tree which can be associated with any persistentobject.

This is a first draft of a tree:

![tree](https://cloud.githubusercontent.com/assets/1849416/16839566/0529f23c-49d0-11e6-9aff-4ca2e2c1ad3b.png)

buehner:
I commited to your branch in #206 
Please have a look at my comment there!

PR #206, opened 2016-07-14
Initialization of tree.

KaiVolland:
This PR is an first approach for a general tree.

The idea is that we can associate any class to a tree node so that we are not limited to layers.

It's a first draft which might be usefull.

The discussion about this tree approach should be made here: #205
You can find an uml class diagramm there, too.


buehner:
I overworked the introduction of the tree modeling of @KaiVolland 

That's what i did:

* Removed the `Tree` and `TreeLeaf` models
  * The `Tree` model is not needed as a tree will always be a (root-) `TreeFolder`
  * The `TreeLeaf` model did not have any additional properties. If you need a simple leaf, just take the `TreeNode` class of my imlementation.
* The `TreeNode` is not abstract in my implementation as this leads to problems regarding hibernate. (And it must not be abstract if we'll use it for leafs!)
  * The `TreeNode` model now has a lot of useful properties.
* I added DAO, Service and web/rest controllers (for `TreeNode`)

One **disadvantage** of the current tree model: There is no generic approach to hold an associated entity (that extends `PersistentObject`) as `PersistentObject` is *NOT* an `@Entity`, but a `@MappedSuperclass`. Generally: The main problem is that in the context of hibernate, there is no easy strategy for an implementation of the classic "composite pattern".

But the current implementation is simple, extensible, working and it represents a tree (if you look at the JSON of a `TreeFolder`).

If you want associated entities on leafs/folders, you'll have to extend the `TreeNode` and/or `TreeFolder` classes as demonstrated here. (I tested this succesfully!):

```Java
@Entity
@Table
public class MySpecialTreeLeaf extends TreeNode {

    @ManyToOne
    private MySpecialEntity entity;

   [...]
}
```
(It's the same for `TreeFolder`)

**Note:** Subclassing will lead to separate tables, but the parent/child association (of all `TreeNode`s and their subclasses) will *always* be stored in the table `TREEFOLDERS_TREENODES` (which is nice in my eyes :wink: ).

<details> 
  <summary>Here is an example JSON of a tree (click to expand)</summary>
   ```Json
{
   "id":43,
   "text":"rootNode",
   "leaf":false,
   "checked":false,
   "expandable":true,
   "expanded":true,
   "children":[
      {
         "id":42,
         "text":"Folder 1",
         "leaf":false,
         "checked":false,
         "expandable":true,
         "expanded":false,
         "children":[
            {
               "id":40,
               "text":"Leaf 3",
               "leaf":true,
               "checked":false,
               "expandable":false,
               "expanded":false
            },
            {
               "id":41,
               "text":"Leaf 4",
               "leaf":true,
               "checked":false,
               "expandable":false,
               "expanded":false
            }
         ]
      },
      {
         "id":38,
         "text":"Leaf 1",
         "leaf":true,
         "checked":false,
         "expandable":false,
         "expanded":false,
         "qTip":"Tooltip Leaf 1",
         "qTitle":"Tooltip Title Leaf 1"
      },
      {
         "id":39,
         "text":"Leaf 2",
         "leaf":true,
         "checked":false,
         "expandable":false,
         "expanded":false
      }
   ]
}
```
</details>

Please review @annarieger @dnlkoch @ahennr @weskamm

**Comment by buehner** [comment]
I added one more commit to enhance the tree modeling (and it's interaction with ExtJS 6 TreeStores and REST):

* Use `@OneToMany` and `@ManyToOne` for the parent/child relation (but not `@ManyToMany`)
  * This commit introduces a property `parentFolder` on the side of the children
  * The child is the owning side here!! (new "parentFolder" property)
* Introduce some useful properties
  * `index` (for ordering inside of a folder)
  * `root` (whether or not a node is a root node/folder)

**Comment by buehner** [comment]
<details> 
  <summary>This is how the JSON of a tree may look like now (click to expand)</summary>
{
   "id":38,
   "text":"rootNode",
   "index":0,
   "root":true,
   "leaf":false,
   "checked":false,
   "expandable":true,
   "expanded":true,
   "children":[
      {
         "id":43,
         "text":"Leaf 4",
         "parentFolder":38,
         "index":0,
         "root":false,
         "leaf":true,
         "checked":false,
         "expandable":false,
         "expanded":false
      },
      {
         "id":39,
         "text":"Folder 1",
         "parentFolder":38,
         "index":1,
         "root":false,
         "leaf":false,
         "checked":false,
         "expandable":true,
         "expanded":false,
         "children":[
            {
               "id":40,
               "text":"Leaf 1",
               "parentFolder":39,
               "index":0,
               "root":false,
               "leaf":true,
               "checked":false,
               "expandable":false,
               "expanded":false,
               "qTip":"Tooltip Leaf 1",
               "qTitle":"Tooltip Title Leaf 1"
            },
            {
               "id":41,
               "text":"Leaf 2",
               "parentFolder":39,
               "index":0,
               "root":false,
               "leaf":true,
               "checked":false,
               "expandable":false,
               "expanded":false
            }
         ]
      },
      {
         "id":42,
         "text":"Leaf 3",
         "parentFolder":38,
         "index":2,
         "root":false,
         "leaf":true,
         "checked":false,
         "expandable":false,
         "expanded":false
      }
   ]
}
</details>

**Note:** In the Java models the `parentFolder` is of type `TreeFolder`, but we configured the Java model to serialize only the IDs (and resolve them vice versa)

**Comment by KaiVolland** [comment]
Very nice! :+1: :heart_eyes:

**Comment by marcjansen** [comment]
What is the state of this PR? /cc @terrestris/owners 

What is missing to get this in?

**Comment by buehner** [comment]
As far as i remember there is nothing missing to get this in. As this are only new files, but not changes on existing code, existing projects should not be affected.

**Comment by marcjansen** [comment]
OK. Looking through the diff right now.

File: src/shogun2-core/src/main/java/de/terrestris/shogun2/model/tree/TreeFolder.java [hunk 1]
```diff
@@ -0,0 +1,138 @@
...
+
+		// folders are usually expandable
+		this.setExpandable(true);
+	}
+
+	/**
+	 *
+	 * @param node
+	 */
+	public void addNode(TreeNode node) {
```
marcjansen: Don't we ususally have models that are POJOs with getters and setters? Why do we have special code here?

src/shogun2-core/src/main/java/de/terrestris/shogun2/model/tree/TreeFolder.java
```diff
@@ -0,0 +1,138 @@
...
+		this.children.add(node);
+	}
+
+	/**
+	 *
+	 * @param node
+	 */
+	public void remove(TreeNode node) {
```
marcjansen: If this is to be kept, it should be called `removeNode`, right?

src/shogun2-core/src/main/java/de/terrestris/shogun2/model/tree/TreeFolder.java
```diff
@@ -0,0 +1,138 @@
...
+	 *
+	 *      According to
+	 *      http://stackoverflow.com/questions/27581/overriding-equals
+	 *      -and-hashcode-in-java it is recommended only to use getter-methods
+	 *      when using ORM like Hibernate
+	 */
+	@Override
+	public int hashCode() {
+		// two randomly chosen prime numbers
+		return new HashCodeBuilder(19, 17)
```
marcjansen: These are the same for `model/module/CompositeModule` I doubt this is a real problem, but we don't have duplicates elsewhere

src/shogun2-core/src/main/java/de/terrestris/shogun2/model/tree/TreeFolder.java
buehner: `TreeFolder` extends `TreeNode`, but requires different values for the `leaf` and `expandable` properties. That's why we set these values in the constructor.

src/shogun2-core/src/main/java/de/terrestris/shogun2/model/tree/TreeNode.java
```diff
@@ -0,0 +1,381 @@
...
+	 * The position of the node inside its parent. When parent has 4 children
+	 * and the node is third amongst them, index will be 2 -> starting with
+	 * index 0
+	 */
+	private int index;
+
+	/**
+	 * True if this is the root node.
+	 */
+	private boolean root;
```
marcjansen: Can the root node be anything else than a `treeFolder`? I.e. dooes this make sense to have at the shared base class?

src/shogun2-core/src/main/java/de/terrestris/shogun2/model/tree/TreeFolder.java
buehner: We can rename this. No problem.

src/shogun2-core/src/main/java/de/terrestris/shogun2/model/tree/TreeNode.java
```diff
@@ -0,0 +1,381 @@
...
+	/**
+	 * Control checkboxes:
+	 * 
+	 * <ul>
+	 *   <li>null: no checkbox will appear</li>
+	 *   <li>true: a checked checkbox will appear</li>
+	 *   <li>false: an unchecked checkbox will appear</li>
+	 * </ul>
+	 */
+	private Boolean checked;
```
marcjansen: A good solution, but directly underneath we have `expandable`. Why don't we have `checkable`? Would that be more consistent?

src/shogun2-core/src/main/java/de/terrestris/shogun2/model/tree/TreeFolder.java
buehner: `CompositeModule` uses `17, 19`. Here we have vice-versa `19, 17`. As far as i know, this should be fine.

src/shogun2-core/src/main/java/de/terrestris/shogun2/model/tree/TreeNode.java
```diff
@@ -0,0 +1,381 @@
...
+	 * @param iconCls the iconCls to set
+	 */
+	public void setIconCls(String iconCls) {
+		this.iconCls = iconCls;
+	}
+
+	/**
+	 * @return the qTip
+	 */
+	public String getqTip() {
```
marcjansen: Is this named correctly? Even though it looks weird, I'd expect it to be called with a capital `Q`, e.g. `getQTip`?

marcjansen:
I have added a couple f minor questions and comments. Once they are adressed this is good to go. Thanks everyone for the work here.

src/shogun2-core/src/main/java/de/terrestris/shogun2/model/tree/TreeNode.java
buehner: I think this is related to the fact that the `NodeInterface` class the EXT framework (which may be considered as the counterpart to our `TreeNode` class), also provides the `root` property for leafs.

Maybe it is possible to move this property to the `TreeFolder` class, but i don't know if this will lead to unexpected behavior.

src/shogun2-core/src/main/java/de/terrestris/shogun2/model/tree/TreeFolder.java
marcjansen: :+1: Fine for me.

src/shogun2-core/src/main/java/de/terrestris/shogun2/model/tree/TreeFolder.java
marcjansen: I am not talking about the constructor but about the methods `addNode` and `remove(Node)`. They are manipulating a childproperty, which isn't POJO anymore. Or is it?

src/shogun2-core/src/main/java/de/terrestris/shogun2/model/tree/TreeFolder.java
marcjansen: As I said above, why do we even have this? It is just a `Collection` right?

src/shogun2-core/src/main/java/de/terrestris/shogun2/model/tree/TreeNode.java
buehner: The ternary differentiation also relies on the fact that EXT handles it this way, i.e. we can easily set a value in our model and EXT will behave as documented/expected.

We could introduce an `expandable` property, but this would require some extra lines of code for a successful interaction with EXT.

src/shogun2-core/src/main/java/de/terrestris/shogun2/model/tree/TreeNode.java
marcjansen: OK. It might make sense to have it initially false at this level then, right?

src/shogun2-core/src/main/java/de/terrestris/shogun2/model/tree/TreeNode.java
marcjansen: OK for me, but it just feels weird if you look at this without thinking about the client framework. leave it as it is.

src/shogun2-core/src/main/java/de/terrestris/shogun2/model/tree/TreeNode.java
buehner: Yes. This is weird. But it is generated (by Eclipse) and it seems to be working. I think this is related to the fact that the property's name is `qTip`.

src/shogun2-core/src/main/java/de/terrestris/shogun2/model/tree/TreeNode.java
buehner: The default value for `boolean` is false in Java, but we can also add this explicitly.

src/shogun2-core/src/main/java/de/terrestris/shogun2/model/tree/TreeFolder.java
buehner: I don't think that we have POJOs as we are extending other classes and we are using annotations (just had a quick look at the POJO definition on wikipedia). When implementing some kind of composite pattern (like here), in my eyes these helper methods (add/remove) are nice to have and did not make any problems yet.

marcjansen: As long as we make a decision I am fine as well. I just don't want things like this to slip in accidentally.

src/shogun2-core/src/main/java/de/terrestris/shogun2/model/tree/TreeFolder.java
marcjansen: Keep it. But rename it please.

src/shogun2-core/src/main/java/de/terrestris/shogun2/model/tree/TreeNode.java
marcjansen: Yes please.

marcjansen:
It all boils down to:

* Give root the `root`the expolicit default value `false`
* Harmonize collection methods

We now have

* `addNode`,
* `remove` and
* `setChildren`

…which is confusing me.

I suggest to name them

* ~`addNode`~ `addChild`,
* ~`remove`~ `removeChild` and
* `setChildren`

You agree?

One final question: 

In `TreeFolder` you write: 

> This is NOT (!) the owning side of the parent/child relation, i.e. if youadd a child to this list and persist the instance of this folder, this will NOT be persisted in the database! You will always have to change the child nodes itself to persist such changes, e.g. by setting a different index (for ordering in a folder) or setting a different parentFolder there.

That same class has the utility methods for adding and removing. Why again do we need them if any changes won't have an effecrt once the TreeFolder is persisted?

src/shogun2-core/src/main/java/de/terrestris/shogun2/model/tree/TreeFolder.java
buehner: I just had another thought: As documented in the JavaDoc, the `TreeFolder` is not the owning side of the parent <-> child relation. So i think the add and remove methods are obsolete as they would not work in a way that calls of theses methods would lead to changes in the database (which would only happen if you set the parent on a child node etc)

So i think i will remove them (as they are never used)

buehner:
@marcjansen We had the same thought regarding "owning side" and the add/remove methods.

I just pushed 2 commits that will remove the add/remove methods and set a default for root.

Feel free to merge

marcjansen:
N1ce. Will merge now.

PR #207, opened 2016-09-05
Revert "Initialization of tree."

buehner:
Reverts terrestris/shogun2#206

We forgot to release a new version of SHOGun2 before merging this. As our release script (currently) only allows a new release based on HEAD, this PR will be reverted and recreated after the release of the new SHOGun2 version.

(We should release a new version before merging this because there is at least one project that has naming conflicts with the new stuff - i.e. it needs adaptions - and should use the (fix) version of the upcoming release)


marcjansen:
Plus 1 for reverting, releasing and then reading the tree stuff. :runner: :dash: 

To be honest, that other project never should have been depending on a unreleased version :wink:

chrismayer:
Also :+1: for reverting, releasing and then reading the tree stuff.

I guess app devs using SHOGun would appreciate a release before major changes.

PR #209, opened 2016-09-08
Use applications as templates for new apps

buehner:
This PR introduces a new property `template` (Boolean) for the Application model. Instances with `template=true` will not be treated as regular applications, e.g. they will not appear if all applications are requested via `rest/applications`.
(Instead, the templates could be queried via `rest/applications/templates` or by their ID `rest/applications/{id}`).

The target of this new "concept" is a simplification of the creation of new applications, e.g.  one now could do the following to create a new app:

1. Query an existing (preconfigured) template (!) application from DB (e.g. there may be different templates for different layouts)
1. Detach/evict the queried instance from the Hibernate Session (by calling the new `evict` method that is also introduced in this PR)
1. Set the ID of the detached object to `null`.
1. Set (or overwrite) all relevant information on the application (like name, modules, map, layers, etc.)
1. Persist the object as usual to create a new one.

Until now, the creation of new webapps was a bit cumbersome because a lot of other entities that are related to an application in its complex hierarchy (like the viewport or other (sub-)modules) had to be handled (i.e. fetched/created and/or persisted) separateley with custom logic before the new application could be persisted. When you did not code that carefully, hibernate quickly complained with exceptions like `org.hibernate.PersistentObjectException: detached entity passed to persist`

### DIFF
--- a/src/shogun2-core/src/main/java/de/terrestris/shogun2/service/AbstractCrudService.java
+++ b/src/shogun2-core/src/main/java/de/terrestris/shogun2/service/AbstractCrudService.java
@@ -12,6 +12,7 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 
 import de.terrestris.shogun2.dao.GenericHibernateDao;
+import de.terrestris.shogun2.helper.IdHelper;
 import de.terrestris.shogun2.model.PersistentObject;
 
 /**
@@ -102,4 +103,26 @@ public abstract class AbstractCrudService<E extends PersistentObject, D extends 
 		dao.delete(e);
 	}
 
+	/**
+	 * Clones an entity by detaching it from the hibernate session and resetting the ID to null.
+	 * The clone will be persisted as a new entity if persist is true.
+	 *
+	 * @param e The entity to clone
+	 * @param persist whether or not the clone should be persisted as a new entity
+	 *
+	 * @throws IllegalAccessException
+	 * @throws NoSuchFieldException
+	 */
+	@PreAuthorize("hasRole(@configHolder.getSuperAdminRoleName()) or hasPermission(#e, 'READ')")
+	public E cloneEntity(E e, boolean persist) throws NoSuchFieldException, IllegalAccessException {
+		dao.evict(e);
+
+		IdHelper.setIdOnPersistentObject(e, null);
+
+		if(persist) {
+			dao.saveOrUpdate(e);
+		}
+
+		return e;
+	}
 }
--- a/src/shogun2-core/src/main/java/de/terrestris/shogun2/dao/ApplicationDao.java
+++ b/src/shogun2-core/src/main/java/de/terrestris/shogun2/dao/ApplicationDao.java
@@ -36,7 +36,7 @@ public class ApplicationDao<E extends Application> extends
 	 */
 	@Override
 	public List<E> findAll() {
-		LOG.trace("Finding all (non-template) applications of ");
+		LOG.trace("Finding all (non-template) applications.");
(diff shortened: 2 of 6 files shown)


src/shogun2-core/src/main/java/de/terrestris/shogun2/service/AbstractCrudService.java
```diff
@@ -102,4 +103,26 @@ public void delete(E e) {
...
+	 * Clones an entity by detaching it from the hibernate session and resetting the ID to null.
+	 * The clone will be persisted as a new entity if persist is true.
+	 *
+	 * @param e The entity to clone
+	 * @param persist whether or not the clone should be persisted as a new entity
+	 *
+	 * @throws IllegalAccessException
+	 * @throws NoSuchFieldException
+	 */
+	@PreAuthorize("hasRole(@configHolder.getSuperAdminRoleName()) or hasPermission(#e, 'READ')")
```
marcjansen: Write permission needed?

src/shogun2-core/src/main/java/de/terrestris/shogun2/service/AbstractCrudService.java
```diff
@@ -102,4 +103,26 @@ public void delete(E e) {
...
+	public E cloneEntity(E e, boolean persist) throws NoSuchFieldException, IllegalAccessException {
+		dao.evict(e);
+
+		IdHelper.setIdOnPersistentObject(e, null);
```
marcjansen: Wasn't this method once only used internally for tests? Now we have apublic method calling it...

marcjansen:
Looking good to me, thanks @buehner. I have added two questions, I'd ask you to answer them and possibly update the PR prior to merging

src/shogun2-core/src/main/java/de/terrestris/shogun2/service/AbstractCrudService.java
buehner: If i want to clone an entity, why should i need a write permission for it? I think it is enough to have the READ permission for an entity to clone it!?

marcjansen: Probably true. But also you have the option to persist the changes. Take what you think makes most sense

src/shogun2-core/src/main/java/de/terrestris/shogun2/service/AbstractCrudService.java
buehner: Yes. We could also introduce a `setId()` method on the `PersistentObject` model, which would facilitate misuse. I am happy with the helper method (which - of course - should only be used if you know what you do).

src/shogun2-core/src/main/java/de/terrestris/shogun2/service/AbstractCrudService.java
buehner: Yes. If i want to persists, i will persist the clone, but the original entity will not be touched, so i think the READ permission should be enough here.

src/shogun2-core/src/main/java/de/terrestris/shogun2/service/AbstractCrudService.java
marcjansen: Good feedback. Keep it as it is and merge at will

buehner:
The template approach proposed in this PR unfortunately has some drawbacks. The main problem is that a lot of hibernate magic is happening (like auto flushing at the end of transactions etc), which makes this approach not practicable.

So i would like to drop this PR without merging.

marcjansen:
Fine for me. Let's close this.

PR #210, opened 2016-09-20
Send error message to client when PW reset failed

buehner:
Just a minor change to send a more meaningful error message to the client if a password reset did not work as expected.


buehner:
any comments @marcjansen @dnlkoch @KaiVolland @annarieger @weskamm ?

or can this be merged?

PR #211, opened 2016-09-21
Switch connection pooling to HikariCP

ahennr:
By this PR the connection pooling is changed from [c3p0](http://www.mchange.com/projects/c3p0/) to [HikariCP](http://brettwooldridge.github.io/HikariCP/).

Beside that [HikariCP outperforms](https://github.com/brettwooldridge/HikariCP-benchmark) c3p0 insofar as performance is concerned, the major advantage is that there are less (required) parameters to tweak which makes the configuration of the connection pooling much simpler.

In Maven - `pom.xml`: As the `artifactId` and version differs depending on the JDK version (`HikariCP-java7` for Java7 resp. `HikariCP` for Java8), I needed to introduce them as dependencies into the profiles section.


buehner:
LGTM, but i think a second review from @marcjansen @dnlkoch @KaiVolland @annarieger @weskamm or @chrismayer would be nice before merging this...

buehner:
There's one thing that came to my mind: We should update the release notes and make a new release before merging this!

@ahennr Maybe you can update the Release notes, if you find time for that!? :wink:

buehner:
Thx @ahennr for adding the release notes. I will merge now

marcjansen:
👏👏👏👏

PR #212, opened 2016-09-28
HttpUtil - use Credentials interface of apache httpcomponents

ahennr:
This PR replaces the `String` parameters for username and password passed to the `BasicCredentialsProvider` in `send(...)` method by the [`Credentials`](https://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/auth/Credentials.html)  interface.

For example, for simple username and password credentials, the implementing class `UsernamePasswordCredentials` can be used.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -2,6 +2,8 @@
 
 ## 0.1.1 (xxxx-xx-xx)
 
+* Changes:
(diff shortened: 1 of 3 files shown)


buehner:
LGTM! I'll merge soon if nobody else has comments...

ahennr:
Alright, I added API methods with the "old" signature and their corresponding tests (see 6ab9a67) to ensure backwards compatibility to older versions of SHOGun2. The release note were updated as well.

marcjansen:
:+1:

buehner:
Thx @ahennr  for adding the "old" signature methods again. By doing this, we assure backwards compatibility and there is no reason to make a release before merging this (as we only "extend" the existing functionality). So i'll merge now...

PR #213, opened 2016-10-14
Introduce a CSV2ExtJsLocale interface

buehner:
This PR introduces a new interface that can be used to easily retrieve ExtJS compatible locale JSONs that are generated on the basis of a CSV file with all the locale information (for different languages).

One of the main advantages of this approach is that a CSV can easily be used in Excel or LibreOffice to manage the translation - even by non-technical persons. Another advantage is the extensibility for new languages (which would simply be additional columns in the CSV).

The retrieved JSON is compatible with the BasiGX language combo: https://github.com/terrestris/BasiGX/blob/master/src/view/combo/Language.js

Example:

A CSV file `src/main/resources/META-INF/locale/client.csv` with the following content:

```
component;field;de;en
MyProject.some.Component;config.data.title;"de:title;and something after a semicolon";en:foobar
MyProject.some.Component;config.data.tooltip;de:tooltip;en:tooltip
MyProject.some.Component;config.data.dragText;{0} Zeile(n) ausgewählt;{0} line(s) selected
MyProject.another.Component;justonefield;de:foo;en:foo
MyProject.another.Component;data.a;de:data.a;en:data.a
MyProject.another.Component;data.b;de:data.b;en:data.b
MyProject.yet.another.Component;config.myArray[];de:v1,v2,v3, ...;en:v1,v2,v3, ...
```

would be translated to

```JSON
{
   "MyProject.another.Component":{
      "data":{
         "a":"de:data.a",
         "b":"de:data.b"
      },
      "justonefield":"de:foo"
   },
   "MyProject.some.Component":{
      "config":{
         "data":{
            "title":"de:title;and something after a semicolon",
            "tooltip":"de:tooltip",
            "dragText":"{0} Zeile(n) ausgewählt"
         }
      }
   },
   "MyProject.yet.another.Component":{
      "config":{
         "myArray":[
            "de:v1",
            "v2",
            "v3",
            " ..."
         ]
      }
   }
}
```

when the following web-interface is called: `http://localhost:8080/yourProject/locale/client/de.json`

To retrieve the english version (i.e. the values from the `en` column in the CSV), you would simply have to replace the `de` in the URL above with `en`. You could also retrieve the content of other CSV locale files by adapting `client` in the above URL to the name of your CSV (without `.csv`), e.g. `http://localhost:8080/yourProject/locale/admin/en.json` for the english version of a CSV named `admin.csv`.

To easily generate the needed `Map<String, Object>` (map of maps) in Java, which will be translated by the Jackson lib to an according JSON, the very useful `StructureUtils.putValueIntoMapOfMaps` method of the JBoss/elasticsearch project is used (as this method supports the dot notation in the second column of the CSVs).

Please review this first version of the locale interface. Thanks!


buehner:
Just to keep this in mind: I still have to check if the current webapp archetype is compatible to this new feature. It is possible that some adaptions in the `web.xml` are necessary (to allow `.json` next to `.action in the URL).

src/shogun2-core/src/main/java/de/terrestris/shogun2/service/Csv2ExtJsLocaleService.java
```diff
@@ -0,0 +1,144 @@
...
+	@Autowired
+	private ResourceLoader resourceLoader;
+
+	/**
+	 *
+	 * @param locale
+	 * @param locale2
+	 * @return
+	 * @throws Exception
+	 */
```
buehner: JavaDoc is not correct here...

src/shogun2-core/src/main/java/de/terrestris/shogun2/service/Csv2ExtJsLocaleService.java
```diff
@@ -0,0 +1,144 @@
...
+		}
+
+		Reader reader = new FileReader(csvResource.getFile());
+		CSVReader csvReader = new CSVReader(reader, ';', '"', '\\');
+
+		try {
+			int columnIndexOfLocale = detectColumnIndexOfLocale(locale, csvReader);
+
+			List<String> nextLine;
+			while ((nextLine = Arrays.asList(ArrayUtils.nullToEmpty(csvReader.readNext()))).isEmpty() == false) {
```
marcjansen: You're kidding me, right? 😆 IMO this line is too long and complex.

src/shogun2-core/src/main/java/de/terrestris/shogun2/service/Csv2ExtJsLocaleService.java
```diff
@@ -0,0 +1,144 @@
...
+
+				if (resultMap.containsKey(component)) {
+					componentEntry = resultMap.get(component);
+				} else {
+					componentEntry = new TreeMap<String, Object>();
+				}
+
+				// handle arrays
+				if (field.contains("[]")) {
+					// assure that [] occurs only once at end of string
```
marcjansen: You cannot assume that, right? An arra might contain objects.

src/shogun2-core/src/main/java/de/terrestris/shogun2/service/Csv2ExtJsLocaleService.java
```diff
@@ -0,0 +1,144 @@
...
+		if (headerLine == null || headerLine.isEmpty()) {
+			throw new Exception("CSV locale file seems to be empty.");
+		}
+
+		if (headerLine.size() < 3) {
+			// we expect at least three columns: component;field;locale1
+			throw new Exception("CSV locale file is invalid: Not enough columns.");
+		}
+
+		// start with the third column as the first to columns must not be a
```
marcjansen: ...first *two* columns...

marcjansen:
One thing I forgot. Can you please add tests? This looks like it can be done easily.

src/shogun2-core/src/main/java/de/terrestris/shogun2/service/Csv2ExtJsLocaleService.java
buehner: :grin: :+1:

src/shogun2-core/src/main/java/de/terrestris/shogun2/service/Csv2ExtJsLocaleService.java
buehner: No. I am not kidding as the csvReader does not provide a `hasNext()` method... :wink: 
But i'll try to simplify this in another way...

buehner:
I addressed the mentioned issues and added tests.
But do me a favour @marcjansen : Don't say that it looks easy to create tests in a spring mvc setup :laughing: :joy: 

I am also not 100% sure whether this code should go to SHOGun2, but as you said: it does not really hurt. What i would like to have in future is some kind of plugin mechanism to pull in such features as optional ones.

PR #215, opened 2016-10-28
Extend button model and add dao/service/controller

buehner:
This commit adds a new Boolean property to the button model: `defaultButton`
The value of this property defines whether a button is a default button in an Application or not (i.e. an optional/configurable one).

Within this PR the missing DAO/Service/Controller are also added.


ahennr:
:+1: from me.

annarieger:
also from me :+1:

buehner:
i dont think that a release is needed as v0.1.1. has just been released two days ago.
the hikari change introduced in #211 is more complex to adapt than the introduction in this PR

thx. i'll merge now

PR #216, opened 2016-10-28
Introduce a simple rest filter

buehner:
This PR introduces a new rest GET interface `/rest/{myEntity}/filter` which allows basic/simple filtering based on primitve fields of (arbitrary) entities.

Examples:
- `/rest/modules/filter?name=test`
  - would return all modules with module.name = 'test'
- `/rest/modules/filter?id=17&name=mymodule`
  - would be translated to a filter `id == 17 AND name =='mymodule'`
- `/rest/modules/filter?name=n1&name=n2&xTYpe=foo`
  - would be translated to a filter `(name == 'n1' OR name == 'n2') AND (xtype == 'foo')'`

This will only provide basic filtering possibilities as this currently only works on simple properties of entities.

In case of boolean fields, the values in the query string may be `true` or `1` (or `false`/`0`).

If you try to filter on attribute names that dont exist in the target entity, they will be ignored!

Attribute names in the query string may be case insensitive (currently with one known exception: if you use a fieldname multiple times in the query string, the field name should always be exactly the same within the query string. so the following query string would not lead to an expected result currently: `filter?NAme=n1&naME=n2&xtype=foo`)

Tests for this PR may follow in the next week :wink:

### DIFF
--- a/src/shogun2-core/src/main/java/de/terrestris/shogun2/rest/AbstractRestController.java
+++ b/src/shogun2-core/src/main/java/de/terrestris/shogun2/rest/AbstractRestController.java
@@ -10,10 +10,12 @@ import org.apache.commons.lang3.StringUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
+import org.springframework.util.MultiValueMap;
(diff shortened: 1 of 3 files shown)


dnlkoch:
Great! +1

marcjansen:
Merging this now.

PR #218, opened 2016-11-14
 Add HTTP util methods with custom HTTP headers

chrismayer:
This adds methods to the ``HttpUtil`` class offering the possibility to add custom HTTP headers to the requests.

Would be great if the original devs of the ``HttpUtil`` class can have a look at this and if they agree in general I would finalize this by adding some tests and API docs.

### DIFF
--- a/src/shogun2-core/src/main/java/de/terrestris/shogun2/util/http/HttpUtil.java
+++ b/src/shogun2-core/src/main/java/de/terrestris/shogun2/util/http/HttpUtil.java
@@ -83,7 +83,24 @@ public class HttpUtil {
 	 * @throws HttpException
 	 */
 	public static Response get(String url) throws URISyntaxException, HttpException {
-		return send(new HttpGet(url), null);
+		return send(new HttpGet(url), null, null);
+	}
+
+	/**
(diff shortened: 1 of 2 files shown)


marcjansen:
I am not opposed to this and would be plus 1 for merging (after comments and tests are added).

With this in, we will have > 90 static public methods in this utility class. Basically all of them are overloaded.  Looking through the diff was already quirte hard. Is there anything we can do to have this less verbose? 

@buehner, do you want to add something?

buehner:
I am not one of the original devs of this class (--> @ahennr and @dnlkoch) but i checked out the `http-util-headers` branch and could successfully use it.
As the changes seem to be backwards-compatible (through overloading!?), i dont have a problem to merge this. If there are tests missing, it would be nice to add them.

ahennr:
:+1: from me as well. As @buehner mentioned above, the addition of missing tests would be fine.

chrismayer:
Thank you all for your feedback. I just finalized this and added tests.

buehner:
I just checked out your changes and tested them with different existing projects to assure backwards compatability.

As everything is fine, i'll merge now!

Thx for the work @chrismayer

PR #219, opened 2016-11-29
Remove unused auth provider

buehner:
This removes a definition of the `AuthenticationProvider`, which is *never* used in the `ContentInitializer`, but may lead to problems when you do not have a bean of the SHOGun2 auth provider (but some other kind of authentication like pre-auth by http header info).

Please merge if you dont have any concerns!


marcjansen:
LGTM, please merge.

buehner:
thx :+1:

PR #220, opened 2016-12-06
Extended LayerAppearance model

dnlkoch:
This adds two new fields `hoverable` and `hoverTemplate` to the `LayerAppearance` model.


buehner:
:+1: from me

PR #221, opened 2016-12-06
Force UTF8 encoding when reading locale csv

buehner:
The locale CSV is encoded in UTF8. The current code only worked on systems where UTF8 was the default file encoding (as `FileReader` always uses the system default). The code did not work properly on systems with a different default encoding, e.g. `file.encoding=ANSI_X3.4-1968`.

This PR fixes this problem by reading the locale CSV with UTF8 encoding.


dnlkoch:
:+1:

PR #223, opened 2016-12-12
Added 'properties' field to layer appearance model

annarieger:
Added hash map for properties to the layer appearance model.

As the `properties` will be defined as `HashMap` object, this can contain a list of key-value pairs that will be properly resolved in frontend.

Please review @dnlkoch @marcjansen


marcjansen:
Thanks!. I am `+1`, but would favor the opinion of @dnlkoch or @buehner.

buehner:
:+1:

PR #228, opened 2017-01-30
Make the presence of role information optional

buehner:
The framework was very rigid regarding the presence of role information coming from the configuration. But as soon as role information is coming dynamically from external sources, this makes it hard to get rid of this (in those cases unnecessary/useless) configuration.

This PR simplifies this by making the presence of configurable role information optional.


buehner:
Currently those values are defaulting to the empty string "". You mean something like "ROLE_UNDEFINED" or simply "UNDEFINED"?

PR #229, opened 2017-02-07
Make use of plain principal object in permission evaluation configurable

buehner:
The `Shogun2PermissionEvaluator` always fetched the full user object from the database. But in cases where your `Principal` from the spring security context (which usually is a SHOGun2 user instance) holds temporary information that is not persisted in the database but required for the permission evaluation, you may want to use the plain principal object from the security context instead of retrieving the full user from the database.

This PR makes this configurable (i.e. you now can set the `usePlainPrincipal` property to `true` in your security XML).

Existing projects should not be affected by this change as the default for this value is `false`, which leads to the same behavior as before.

### DIFF
--- a/src/shogun2-core/src/main/java/de/terrestris/shogun2/security/access/Shogun2PermissionEvaluator.java
+++ b/src/shogun2-core/src/main/java/de/terrestris/shogun2/security/access/Shogun2PermissionEvaluator.java
@@ -77,8 +77,10 @@ public class Shogun2PermissionEvaluator implements PermissionEvaluator {
 
 			User user = null;
 
-			if(authentication.getPrincipal() instanceof User) {
-				final User principal = (User) authentication.getPrincipal();
+			final Object principalObject = authentication.getPrincipal();
+
+			if(principalObject instanceof User) {
+				final User principal = (User) principalObject;
 
 				if(usePlainPrincipal == true) {
 					user = principal;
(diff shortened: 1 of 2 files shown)


buehner:
I added some tests for the new case...

marcjansen:
Please merge at will.

weskamm:
Have you thought about some problematic cases, e.g. when the principal is not in sync with the information persisted in the database? E.g. changing roles of a user and calling saveOrUpdate, does this always update the principal correctly? I remember having trouble with theses cases...
If some information that are contained in both (db and principal) may get out of sync, you should add a big warning text to your new flag

buehner:
@weskamm Unless your application cares about this, the principal will never be in sync with the information in the database when something should be changed on the user object after authenticating. Even in the default SHOGun2-implementation a call of saveOrUpdate on the user would not update the principal. So in my eyes the problem you describe also exists for the default/previous implemantation (when the new flag is false).

When calling the `getUserBySession` from the user service you'll get a fresh version of the user data based on the principal's ID. In our case, the principal is **not** a *database-connected* user object (but simply created with `new User()`). Our user object holds temporary information (like email, roles etc.) coming from a different data source, which we don't want/are not allowed to persist in the shogun2 database. But we still want to rely on this information during permission evaluation. So the new flag only allows you to decide whether you want the principal object to be shogun2-database-dependent or whether the plain principal object should be used in permission evaluation (whatever has been set as principal during authentication).

PR #231, opened 2017-02-21
Use system JDK when compiling shogun2

buehner:
Just a minor change.

When compiling/building SHOGun2, the system JDK will be used (as `jdk.version`) is set in a profile based on the JDK in use...


ahennr:
:top: nice one

PR #233, opened 2017-02-22
Fix JPA annotation to also work in postgres

buehner:
The JPA annotation `@Column(length = Integer.MAX_VALUE)` (ping @dnlkoch) should only be used in case of `byte[]`, but not for `String` as postgres will not be able to create a varchar column with that size.

Using the `@Lob` annotation for `String` works fine for H2 and Postgres. In case of Postgres, a column of type `TEXT` will be created.


dnlkoch:
Aaaaaaaand SHOGun2 is great again! Thanks @buehner!

PR #235, opened 2017-02-24
Allow configurable sparse output of REST GET requests

marcjansen:
This adds the possibility to optionally get a sparse result from REST GET calls against the abstract REST controller. This can be usefull if you are not interested in the complete representation of objects, but only certain fields are important for you.

In order to only get e.g. the fields `name` and `id` of all entities `TestEntity` (mapped to `/rest/testentity`): the following request can be used:

    /rest/testentity?output:only=name,id

All other fields of the returned objects will be `null`.

This can be combined with the existing `filter`-API:

    /rest/testentity/filter?id=42&output:only=name,id

The above will give you back the sparse `TestEntity` serialisation where only the fields `name` and `id` have real values, others are `null`.

This implementation is backwards compatible for the REST API.

I didn't find time to add tests for now, but wanted to get feedback fast; so please review.

### DIFF
--- a/src/shogun2-core/src/main/java/de/terrestris/shogun2/dao/GenericHibernateDao.java
+++ b/src/shogun2-core/src/main/java/de/terrestris/shogun2/dao/GenericHibernateDao.java
@@ -13,10 +13,13 @@ import org.hibernate.Session;
 import org.hibernate.SessionFactory;
 import org.hibernate.criterion.Criterion;
 import org.hibernate.criterion.Order;
+import org.hibernate.criterion.ProjectionList;
 import org.hibernate.criterion.Projections;
+import org.hibernate.criterion.PropertyProjection;
 import org.hibernate.criterion.Restrictions;
 import org.hibernate.proxy.HibernateProxy;
 import org.hibernate.transform.DistinctRootEntityResultTransformer;
+import org.hibernate.transform.Transformers;
 import org.joda.time.DateTime;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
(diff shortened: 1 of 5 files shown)


buehner:
Looks like a really nice feature @marcjansen .
Some tests would be fine, but for my part, this could already be merged...

marcjansen:
Thanks for the review. I would love to have the name of the parameter be configurable,  though.

I also like the word sparse better than the other ones. I will continue from here anytime soon.

weskamm:
:+1:

buehner:
ping @marcjansen 
There are conflicts you have to resolve before we can merge :wink:

buehner:
ping @marcjansen :poodle:

buehner:
Is it possible to resolve the conflicts @marcjansen ?

marcjansen:
The conflicts can be resolved for sure, but I am not sure I want this functionality in the way it is implemented here. We can also simply close this and revive this later if we want to. You decide.

buehner:
Hey @terrestris/devs 

as the changes within this PR are really "nice to have", i'd like to re-initiate a quick discussion on this topic. In my eyes, there is nothing wrong with merging this -> :+1: 

It might be nice to have such features available somedays, even though there is no "must-have"-requirement at the moment.

Any opinions?

I think we need a decision before the whole thing lingers for another year.

hwbllmnn:
Agreed. We just have to fix the merge conflict first...

marcjansen:
Merge conflicts should be gone, please double check if this does what it tells it does. Do with this PR whatever you want.

buehner:
Thanks for the feedback. I'll checkout the branch and check if everything is still working as expected.

buehner:
OK. I got this running succesfully. There are a few drawbacks that i will document here. But as this is just an enhancement and will not break any existing implementation, i'll merge afterwards.

Known issues of this new feature:

* This only works for properties of primitve type (in case a non-primitve property is requested, the complete object will be serialized "normally")
* Properties that are not requested, still appear in the output with `null` values

Thanks for this @marcjansen !
Maybe we'll find some time to enhance this in future.

PR #236, opened 2017-03-01
Fix postgres transactional error when using @Lob

buehner:
The use of the `@Lob` annotation at a (possibly large) string field lead to errors when using postgres:

`org.postgresql.util.PSQLException: Large Objects may not be used in auto-commit mode.`

This change will fix the problem...


annarieger:
Thanks for fixing it :pineapple: 

Please merge when green

PR #237, opened 2017-03-01
Clone/Copy persisted trees

buehner:
This PR allows you to clone existing `TreeNode`s, i.e. if you have a persisted tree (root) node, all child nodes and the tree/root node itself would be initialized, detached from the hibernate session and re-persisted as a new entry in the database when the new method `cloneAndPersistTreeNode` is called.

This is helpful if you want to copy/clone an `Application` with the underlying layer tree. If you would simply set the original root node of the layer tree for the copy-application, the same tree would be used.

With the changes of this PR, the whole tree can be copied/cloned.

For this, two new methods in the `GenericHibernateDao` have been introduced and are used within the new `cloneAndPersistTreeNode` method:

* `unproxy(E e)`, which eagerly initializes an entity (just to be sure that all data that would be fetched lazily is available)
* `evict(E e)` to detach an entity from the hibernate session, which allows us to set the ID to null afterwards, so that a call of `saveOrUpdate` will create a new database entry.

Please review and merge. Ping @marcjansen @annarieger

### DIFF
--- a/src/shogun2-core/src/main/java/de/terrestris/shogun2/dao/GenericHibernateDao.java
+++ b/src/shogun2-core/src/main/java/de/terrestris/shogun2/dao/GenericHibernateDao.java
@@ -7,12 +7,14 @@ import java.util.Map;
 
 import org.apache.log4j.Logger;
 import org.hibernate.Criteria;
+import org.hibernate.Hibernate;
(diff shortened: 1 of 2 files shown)


weskamm:
Looking good to me too, merging now...

PR #238, opened 2017-03-06
Forward deletion of plugins to applications

marcjansen:
This ensures that when a plugin is deleted, references to that plugin
from within applications are removed as well.

To achieve this, we override the generic `delete`-method and update apps
using a new autowired application service.

The code contains some `TODO` blocks which might be interesting to resolve
in the future.

Please review.

### DIFF
--- a/src/shogun2-core/src/main/java/de/terrestris/shogun2/service/PluginService.java
+++ b/src/shogun2-core/src/main/java/de/terrestris/shogun2/service/PluginService.java
@@ -36,7 +36,9 @@ public class PluginService<E extends Plugin, D extends PluginDao<E>> extends
 	/**
 	 * The application service which we e.g. need when plugins are deleted.
 	 */
-	private ApplicationService<Application, ApplicationDao<Application>> applicationService = null;
+	@Autowired
+	@Qualifier("applicationService")
+	private ApplicationService<Application, ApplicationDao<Application>> applicationService;
 
 	/**
 	 * Default constructor, which calls the type-constructor
@@ -103,34 +105,29 @@ public class PluginService<E extends Plugin, D extends PluginDao<E>> extends
 	@Override
 	@PreAuthorize("hasRole(@configHolder.getSuperAdminRoleName()) or hasPermission(#plugin, 'DELETE')")
 	public void delete(E plugin) {
-		ApplicationService<Application, ApplicationDao<Application>> appService = this.applicationService;
-		if (appService == null) {
+		if (applicationService == null) {
 			LOG.error("Plugin cannot be deleted, failed to autowire application service");
 			return;
 		}
 		// TODO We should have a more elegant way of finding the affected applications instead of fetching them all
 		//      Maybe a generic method `service.findAllHavingSubentity(String nameOfCollPropToCheck, Class subtype)`
 		//      in the AbstractCrudService?
-		List<Application> apps = appService.findAll();
+		List<Application> apps = applicationService.findAll();
 		Integer pluginId = plugin.getId();
 		for (Application app : apps) {
 			List<Plugin> plugins = app.getPlugins();
 			if (plugins != null && plugins.contains(plugin)) {
-				if (LOG.isDebugEnabled()) {
-					String msg = String.format(
-						"Remove plugin (id=%s) from application (id=%s)",
-						pluginId, app.getId()
-					);
-					LOG.debug(msg);
-				}
+				String msg = String.format(
+					"Remove plugin (id=%s) from application (id=%s)",
+					pluginId, app.getId()
+				);
+				LOG.debug(msg);
 				plugins.remove(plugin);
 				// TODO will this use the the PreAuthorize annotations of AbstractCrudService wrt WRITE on the app?
-				appService.saveOrUpdate(app);
+				applicationService.saveOrUpdate(app);
 			}
 		}
-		if (LOG.isDebugEnabled()) {
-			LOG.debug(String.format("Delete plugin (id=%s)", pluginId));
-		}
+		LOG.debug(String.format("Delete plugin (id=%s)", pluginId));
 		// call overridden parent to actually delete the entity itself
 		super.delete(plugin);
 	}
@@ -152,7 +149,6 @@ public class PluginService<E extends Plugin, D extends PluginDao<E>> extends
 	 *
 	 * @param applicationService
 	 */
-	@Autowired
 	public void setApplicationService(ApplicationService<Application, ApplicationDao<Application>> applicationService) {
 		this.applicationService = applicationService;
 	}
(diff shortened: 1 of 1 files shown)


src/shogun2-core/src/main/java/de/terrestris/shogun2/service/PluginService.java
```diff
@@ -95,4 +146,14 @@ public String getPluginSource(String simpleClassName) throws Exception {
 	public void setDao(D dao) {
 		this.dao = dao;
 	}
+
+	/**
+	 * Sets the applicationService.
+	 *
+	 * @param applicationService
+	 */
+	@Autowired
```
buehner: We usually have this annotation at the property above...
I think it makes sense to move it there to stay conform with the other services...

src/shogun2-core/src/main/java/de/terrestris/shogun2/service/PluginService.java
```diff
@@ -85,6 +95,47 @@ public String getPluginSource(String simpleClassName) throws Exception {
...
+		}
+		// TODO We should have a more elegant way of finding the affected applications instead of fetching them all
+		//      Maybe a generic method `service.findAllHavingSubentity(String nameOfCollPropToCheck, Class subtype)`
+		//      in the AbstractCrudService?
+		List<Application> apps = appService.findAll();
+		Integer pluginId = plugin.getId();
+		for (Application app : apps) {
+			List<Plugin> plugins = app.getPlugins();
+			if (plugins != null && plugins.contains(plugin)) {
+				if (LOG.isDebugEnabled()) {
```
buehner: I think this check is only blowing up the code...
Why not just use `LOG.debug(...)`!?

src/shogun2-core/src/main/java/de/terrestris/shogun2/service/PluginService.java
buehner: We usually have this annotation at the property above...
I think it makes sense to move it there to stay conform with the other services...

src/shogun2-core/src/main/java/de/terrestris/shogun2/service/PluginService.java
```diff
@@ -29,6 +34,11 @@
 	private String pluginNamespace;
 
 	/**
+	 * The application service which we e.g. need when plugins are deleted.
+	 */
+	private ApplicationService<Application, ApplicationDao<Application>> applicationService = null;
```
buehner: As null is the default init value, this explicit initialization seems to be obsolete

src/shogun2-core/src/main/java/de/terrestris/shogun2/service/PluginService.java
```diff
@@ -85,6 +95,47 @@ public String getPluginSource(String simpleClassName) throws Exception {
...
 	/**
+	 * Removes the passed plugin from all applications and afterwards deletes the plugin itself. This overrides the
+	 * generic method to delete {@link AbstractCrudService#delete(de.terrestris.shogun2.model.PersistentObject)}.
+	 *
+	 * @param plugin
+	 */
+	@Override
+	@PreAuthorize("hasRole(@configHolder.getSuperAdminRoleName()) or hasPermission(#plugin, 'DELETE')")
+	public void delete(E plugin) {
+		ApplicationService<Application, ApplicationDao<Application>> appService = this.applicationService;
```
buehner: Remove this and simply use `applicationService` down below... :wink:

src/shogun2-core/src/main/java/de/terrestris/shogun2/service/PluginService.java
```diff
@@ -85,6 +95,47 @@ public String getPluginSource(String simpleClassName) throws Exception {
...
+				if (LOG.isDebugEnabled()) {
+					String msg = String.format(
+						"Remove plugin (id=%s) from application (id=%s)",
+						pluginId, app.getId()
+					);
+					LOG.debug(msg);
+				}
+				plugins.remove(plugin);
+				// TODO will this use the the PreAuthorize annotations of AbstractCrudService wrt WRITE on the app?
+				appService.saveOrUpdate(app);
```
buehner: I am not 100 % sure, but i think yes as it is a different service. if you would call another secured method from the (plugin) service itself, the annotation of this method would not be respected i think...

src/shogun2-core/src/main/java/de/terrestris/shogun2/service/PluginService.java
buehner: You MUST declare `@Qualifier('applicationService')` here!
Otherwise spring cannot decide which `ApplicationService` to use (if there is an extension like `ProjectApplicationService`) in your project

src/shogun2-core/src/main/java/de/terrestris/shogun2/service/PluginService.java
marcjansen: OK, done.

src/shogun2-core/src/main/java/de/terrestris/shogun2/service/PluginService.java
marcjansen: OK, done.

src/shogun2-core/src/main/java/de/terrestris/shogun2/service/PluginService.java
marcjansen: OK, done.

src/shogun2-core/src/main/java/de/terrestris/shogun2/service/PluginService.java
marcjansen: OK for me.

src/shogun2-core/src/main/java/de/terrestris/shogun2/service/PluginService.java
marcjansen: `setDao(D dao)` (a couple of lines above) has the `@autowired` as well. Is that beacuse it doesn't have the property `dao` itself, but inherits it?

I will change as suggested, though.

marcjansen: I tend to **not** like this, but will obey, as you are the expert and we do it elsewhere as well.

When does it make sense to have two services? If I have `ProjectApplicationService` because of something in my project, I should remove the generic bean for `ApplicationService`, right? We as a frame work shoul have as less dependencies on project XML as possible IMO.

marcjansen:
Hey @buehner I have addressed your remarks in b9894ae, would you like to take another look?

src/shogun2-core/src/main/java/de/terrestris/shogun2/service/PluginService.java
buehner: Yes. Exactly that's the reason

buehner: I dont like it, too, but we dont have another solution right now.

In future, we could try to:

* make the name of the qualifier configurable or
* do not scan more than one service in a hierarchy...

buehner:
Thanks for the code and the comments. I'll merge now.

PR #239, opened 2017-03-06
Add the possibility to find entities that have another entity assigned in a collection

marcjansen:
@buehenr might want to continue from here… :heart:

PR #240, opened 2017-03-06
Add missing security annotation

buehner:
@marcjansen In https://github.com/terrestris/shogun2/pull/239, we forgot an important security annotation!!! Please merge if green


marcjansen:
Very good catch! Will merge when green.

PR #241, opened 2017-03-06
Fix check of collection field

buehner:
This fixes a bug that occures when using the `isCollectionField` method and a concrete element in the collection is a subtype of the element type declaration of the collection...

@marcjansen :rofl:


buehner:
Thx for the tests @marcjansen 
I'll merge now

PR #243, opened 2017-03-08
Clean removal of WpsProcessExecute entities

dnlkoch:
Unlink `WpsProcessExecute` entities from any associated `WpsPlugin` on deletion.


buehner:
What happens if you would not override the `delete` method? (Just asking for interest because `process` is not a collection)

Otherwise: :+1: LGTM

dnlkoch:
Thx for the review @buehner

The default `delete()` method would not unset the assigned `process` entity and in turn the `WpsProcessExecute` could not be deleted because of an `Reference` constraint.

marcjansen:
Post-merge 👍 Thanks

PR #245, opened 2017-03-08
Add description field to layer model

buehner:
This will simply add a description field to the layer model...


annarieger:
Good to go, please merge when green :pineapple:

marcjansen:
Just to be extra sure: will this not throw errors when upgrading shogun and hibernate is in update mode? I.e. is this field nullable?

PR #246, opened 2017-03-14
Introduce a findAllWhereFieldEquals method

buehner:
This introduces a `findAllWhereFieldEquals` method to DAO and service layer, which is very useful and can be used with `PersistentObject`s and primitve types, too.

### DIFF
--- a/src/shogun2-core/src/main/java/de/terrestris/shogun2/dao/GenericHibernateDao.java
+++ b/src/shogun2-core/src/main/java/de/terrestris/shogun2/dao/GenericHibernateDao.java
@@ -110,19 +110,32 @@ public class GenericHibernateDao<E extends PersistentObject, ID extends Serializ
 	public List<E> findAllWhereFieldEquals(String fieldName, Object fieldEntity,
 			Criterion... criterion) {
 
-		final boolean isField = EntityUtil.isField(entityClass, fieldName, true);
+		Class<?> fieldEntityType = null;
+
+		if(fieldEntity != null) {
+			fieldEntityType = fieldEntity.getClass();
+		}
+
+		final boolean isField = EntityUtil.isField(entityClass, fieldName, fieldEntityType, true);
 
 		if (!isField) {
 			String errorMsg = String.format(
-				"There is no field '%s' in the type '%s'",
+				"There is no field '%s' in the type '%s' that accepts instances of '%s'",
 				fieldName,
-				entityClass.getName()
+				entityClass.getName(),
+				fieldEntityType.getName()
 			);
 			throw new IllegalArgumentException(errorMsg);
 		}
 
 		Criteria criteria = createDistinctRootEntityCriteria(criterion);
-		criteria.add(Restrictions.eq(fieldName, fieldEntity));
+
+		if(fieldEntity == null) {
+			criteria.add(Restrictions.isNull(fieldName));
+		} else {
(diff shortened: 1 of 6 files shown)


annarieger:
LGTM!
Could you provide the tests for the new methods?

buehner:
I pushed some commits that add tests for all new methods. On top of that, i added tests for the recently introduced method `findAllWithCollectionContaining`

src/shogun2-core/src/main/java/de/terrestris/shogun2/dao/GenericHibernateDao.java
```diff
@@ -96,6 +96,37 @@ public E findById(ID id) {
...
+			Criterion... criterion) {
+
+		final boolean isField = EntityUtil.isField(entityClass, fieldName, true);
+
+		if (!isField) {
+			String errorMsg = String.format(
+				"There is no field '%s' in the type '%s'",
+				fieldName,
+				entityClass.getName()
+			);
```
marcjansen: Message could be enhanced to read e.g.

```java
String errorMsg = String.format(
	"There is no field '%s' in the type '%s' that accepts instances of '%s'",
	fieldName,
	entityClass.getName(),
	// beware of null, though
	fieldEntity.getClass().getName()
);
```

src/shogun2-core/src/main/java/de/terrestris/shogun2/dao/GenericHibernateDao.java
```diff
@@ -96,6 +96,37 @@ public E findById(ID id) {
...
+	 * as value.
+	 *
+	 * @param fieldName The name of the field
+	 * @param fieldEntity The element that should be set as value
+	 * @param criterion Additional criterions to apply (optional)
+	 *
+	 * @return The list of objects
+	 */
+	@SuppressWarnings("unchecked")
+	public List<E> findAllWhereFieldEquals(String fieldName, Object fieldEntity,
```
marcjansen: What happens if `fieldEntity` is `null`?

I would assume that `isField` becomes `false` and one runs into the `IllegalArgumentException` (at least if you take my suggestion of also checking the type).

But this is a valid question, right? Please find me all instances of `E` where the field named `fieldName` is (or is not) `null`. What do you think?

I think we could branch off to a dfifferent method in the service, but this comes from the top of my head.

src/shogun2-core/src/main/java/de/terrestris/shogun2/util/entity/EntityUtil.java
```diff
@@ -24,6 +24,20 @@
 public class EntityUtil {
 
 	/**
+	 * @param clazz
+	 * @param fieldName
+	 * @param forceAccess
+	 * @return
+	 */
+	public static boolean isField(Class<?> clazz, String fieldName, boolean forceAccess) {
```
marcjansen: Wouldn't it be better to also check for the accepoted type, here or in another method?

Slightly related, we keep passing `forceAccess` around, but on **all**  SHOGun2 model classe we must have it `true`, since our models only have `private` fields. We should probably remove this everywhere, ansd simply pass through `true`. This could be done in another PR of course, and only if you agree.

buehner:
I think i "finalized" this by also handling `null` values. Please review (and merge!?) @marcjansen

marcjansen:
Looking into this now.

marcjansen:
Very nice, Nils. Merging now. Thanks a lot!

PR #249, opened 2017-03-23
Avoid doubled parameters in interceptor post

weskamm:
This removes doubled parameters in an interceptor POST request.
This is open for discussion, i am especially interested in the opinion of @dnlkoch.

The issue i had was an clientside POST with KVPs. This request will be forwarded by the interceptor to the source with KVPs in the BaseUrl and also in the body, leading to doubled KVPs which the GeoServer often does not like.

This change of course would remove the option to have a URL-Get parameter combined with a POST body, but i think this should nearly never be a usecase

### DIFF
--- a/src/shogun2-core/src/main/java/de/terrestris/shogun2/service/GeoServerInterceptorService.java
+++ b/src/shogun2-core/src/main/java/de/terrestris/shogun2/service/GeoServerInterceptorService.java
@@ -475,18 +475,17 @@ public class GeoServerInterceptorService {
 					// parse the content type of the request
 					ContentType contentType = ContentType.parse(request.getContentType());
 
-					// perform the request with the given parameters
+					// perform the POST request to the URI with queryString and with the given body
 					httpResponse = HttpUtil.post(requestUri, body, contentType);
 				} else {
 
-					// perform the request with the given parameters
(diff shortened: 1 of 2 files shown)


marcjansen:
Looking into this now.

marcjansen:
Hey @weskamm. Any chance you can cherry pick https://github.com/marcjansen/shogun2/commit/fafa44935a168a2f34a43b4612e54ecc7418b34f on top of your commit and see if this still works for you?

weskamm:
nice @marcjansen , checked this patch against my current issue and it solves it.

There is still a minor issue:
When sending a POST with query-string parameters (e.g. like foo=bar), it will also be added to the POST body, in case we are posting form data.
To my knowledge there seems to be no problem with this, but it may arise in the future.

I think this is now good 2 go as it fixes my inital problem well.

marcjansen:
Please also cherry pick https://github.com/marcjansen/shogun2/commit/d930f1e

marcjansen:
ping @weskamm

buehner:
ping @weskamm :poodle: :poodle: :poodle: :poodle: :poodle: :poodle: :poodle:

buehner:
thanks @weskamm and @marcjansen 
I'll merge now.

PR #250, opened 2017-03-28
introduced systemPlugin property on Plugin model class

annarieger:
Introduce `systemPlugin` property on `de.terrestris.shogun2.model.Plugin` class.

This additional field can be helpful to determine if we have a system plugin to be able possibly handle it other than a "normal" plugin (e.g. forbid editing or deleting of such instances).

Please review @terrestris/owners

### DIFF
--- a/src/shogun2-core/src/main/java/de/terrestris/shogun2/model/Plugin.java
+++ b/src/shogun2-core/src/main/java/de/terrestris/shogun2/model/Plugin.java
@@ -52,6 +52,13 @@ public class Plugin extends PersistentObject {
 	@Type(type="text")
 	private String styleSheet;
 
+	/**
+	 * Whether the plugin is a system plugin or not.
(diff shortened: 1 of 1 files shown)


src/shogun2-core/src/main/java/de/terrestris/shogun2/model/Plugin.java
```diff
@@ -52,6 +52,12 @@
 	@Type(type="text")
 	private String styleSheet;
 
+	/**
+	 * Whether the plugin is a system plugin or not. The properties and
+	 * source code of system plugins cannot be edited.
+	 */
```
marcjansen: What about the following API  docs:

This can e.g. be used to disallow certain actions like editing of the source code. No such limitation will be enforced from SHOGun itself.

annarieger:
Thanks for suggestion @marcjansen, APIs are adjusted :pineapple: 

Merge it after travis is green.

PR #251, opened 2017-04-03
Save default values on wps process

buehner:
Do not save default values on a `WpsPrimitve` (which does not allow to the re-use of equal instances with different default values).

Instead, we'll store default values for input params on instances of `WpsProcessExecute`


marcjansen:
Looks good to me. Please merge

PR #252, opened 2017-04-06
Adjust access modifiers

dnlkoch:
Within this PR a set of access modifiers are adjusted to allow public access and/or extension in subclasses.


marcjansen:
💪🖒☝

PR #254, opened 2017-04-10
Add getPathInDocumentAsNodeList()

dnlkoch:
This adds a new method `getPathInDocumentAsNodeList()` returning a XPath match as `NodeList` (instead of a `String` in getPathInDocument). This may be useful, if the XPath query returns a result, that is not capable of being serialized by the use of the built-in `toString()` method (e.g. a list of `Node`s).

Currently not used in SHOGun2 itself.


ahennr:
LGTM :+1:, plz merge.

PR #256, opened 2017-04-27
Use UTF-8 charset as default/fallback in interceptor

buehner:
If charset cannot be parsed from the content type of a request (e.g. if content type is only `text/xml` but not `text/xml; charset=utf-8`), we should use UTF-8 as the default/fallback.


dnlkoch:
LGTM!

weskamm:
+1 from me!

PR #257, opened 2017-05-10
Simplify file upload

buehner:
This is a simplification of the file upload. When using the Ext framework with a fileupload field and form.submit(), the response content type had to be text/html in the past, so a result json was sent as a string (and parsed in the client again).

It seems that it is now possible to return the json with the correct content type (application/json), which allows a simplification of the code.

### DIFF
--- a/src/shogun2-core/src/main/java/de/terrestris/shogun2/web/FileController.java
+++ b/src/shogun2-core/src/main/java/de/terrestris/shogun2/web/FileController.java
@@ -16,14 +16,10 @@ import org.springframework.web.bind.annotation.RequestMethod;
 import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.multipart.MultipartFile;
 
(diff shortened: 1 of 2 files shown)


marcjansen:
```
Results :
Failed tests:   uploadFile_shouldWorkAsExpected(de.terrestris.shogun2.web.FileControllerTest): Content type expected:<text/html> but was:<application/json;charset=UTF-8>
  uploadFile_shouldReturnErrorMessageIfUploadIsEmpty(de.terrestris.shogun2.web.FileControllerTest): Content type expected:<text/html> but was:<application/json;charset=UTF-8>
Tests run: 685, Failures: 2,
```

buehner:
I adapted the test to check for the changed content type

marcjansen:
:+1: for me, any objections @terrestris/owners ?

weskamm:
I tested this change against my current shogun2 project and it works very well, so you get a +1 from me

PR #258, opened 2017-05-12
Add field resolution to MapConfig

dnlkoch:
This adds the field `resolution` to the `MapConfig` model where it can be used to persist the initial/current resolution of a map.


KaiVolland:
LGTM. Please merge.

buehner:
:+1: :poodle:

PR #260, opened 2017-07-11
Precheck plugin deletion

buehner:
Introduce a new web interface that allows to check from which applications a plugin would be removed in case of deletion.

### DIFF
--- a/src/shogun2-core/src/main/java/de/terrestris/shogun2/service/PluginService.java
+++ b/src/shogun2-core/src/main/java/de/terrestris/shogun2/service/PluginService.java
@@ -102,7 +102,7 @@ public class PluginService<E extends Plugin, D extends PluginDao<E>> extends
 	 * @param pluginId
 	 * @return List of application names that contain the given plugin
 	 */
-	@PreAuthorize("hasRole(@configHolder.getSuperAdminRoleName()) or hasPermission(#layerId, 'de.terrestris.shogun2.model.Plugin', 'DELETE')")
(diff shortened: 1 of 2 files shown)


buehner:
Once https://github.com/terrestris/shogun2/pull/262 is merged, these changes can be merged here to make the build green

buehner:
As this is green now, please review @marcjansen @dnlkoch

src/shogun2-core/src/main/java/de/terrestris/shogun2/service/PluginService.java
```diff
@@ -97,6 +98,30 @@ public String getPluginSource(String simpleClassName) throws Exception {
 	}
 
 	/**
+	 *
+	 * @param pluginId
+	 * @return List of application names that contain the given plugin
+	 */
+	@PreAuthorize("hasRole(@configHolder.getSuperAdminRoleName()) or hasPermission(#layerId, 'de.terrestris.shogun2.model.Plugin', 'DELETE')")
```
dnlkoch: `#layerId` -> `#pluginId`?

dnlkoch:
One minor remark only. Feel free to merge!

src/shogun2-core/src/main/java/de/terrestris/shogun2/service/PluginService.java
buehner: nice catch! i'll fix it

PR #262, opened 2017-07-12
Dont test archetype to fix build

buehner:
After investing several hours to find out why the latest builds are not working anymore - even in case of no changes: https://github.com/terrestris/shogun2/pull/261 - the only solution to fix the builds currently seems to NOT test the webapp archetype.

For some reason the maven on the travis machine is not able to access the archetype from the local repository (even though it has just been installed there one step before).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,21 +22,6 @@ before_script:
 
 script:
   - mvn --batch-mode --threads 2.0C --show-version clean test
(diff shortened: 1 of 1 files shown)


buehner:
This will fix https://github.com/terrestris/shogun2/pull/260

marcjansen:
Please merge. Thanks.

PR #265, opened 2017-07-26
Add endpoint to retrieve all request mappings as JSON

chrismayer:
This introduces an HTTP-endpoint ``/endpointdoc`` which delivers all mapped HTTP endpoints as JSON objects. Hereby all instances (types and methods) annotated with ``@RequestMapping`` annotations in ``@Controller`` classes are analyzed.

This is very useful to build an automated API documentation for all HTTP endpoints.

Sample output:

``` javascript
{
    name : null,
    patternsCondition : {
        patterns : [ "/my-super-shogun-endpoint" ],
        empty : false
    },
    methodsCondition : {
        methods : [ "POST", "PUT" ],
        empty : false
    },
    paramsCondition : {
        expressions : [],
        empty : true
    },
    headersCondition : {
        expressions : [],
        empty : true
    },
    consumesCondition : {
        expressions : [ {
            mediaType : {
                type : "application",
                subtype : "json",
                parameters : {},
                qualityValue : 1,
                charSet : null,
                wildcardType : false,
                concrete : true,
                wildcardSubtype : false
            },
            negated : false
        } ],
        empty : false,
        consumableMediaTypes : [ {
            type : "application",
            subtype : "json",
            parameters : {},
            qualityValue : 1,
            charSet : null,
            wildcardType : false,
            concrete : true,
            wildcardSubtype : false
        } ]
    },
    producesCondition : {
        expressions : [],
        empty : true,
        producibleMediaTypes : []
    },
    customCondition : null
}
```

### DIFF
--- a/src/shogun2-core/src/main/java/de/terrestris/shogun2/service/EndpointDocService.java
+++ b/src/shogun2-core/src/main/java/de/terrestris/shogun2/service/EndpointDocService.java
@@ -1,0 +1,38 @@
+package de.terrestris.shogun2.service;
+
+import java.util.Set;
+
+import org.springframework.security.access.prepost.PreAuthorize;
+import org.springframework.stereotype.Service;
+import org.springframework.web.servlet.mvc.method.RequestMappingInfo;
+import org.springframework.web.servlet.mvc.method.annotation.RequestMappingHandlerMapping;
+
+import de.terrestris.shogun2.web.EndpointDocController;
+
+/**
+ * Service class for the {@link EndpointDocController}.
+ *
+ * @author Christian Mayer
+ */
+@Service("endpoinDocService")
+public class EndpointDocService {
+
+	/**
+	 * Returns all RequestMappingInfo instances from type and
+	 * method-level @RequestMapping annotations in @Controller classes.
+	 *
+	 * @param requestMappingHandlerMapping
+	 *            The RequestMappingInfo collection of Spring
+	 * @return A set of RequestMappingInfo
+	 */
+	@PreAuthorize("hasRole(@configHolder.getSuperAdminRoleName())")
+	public Set<RequestMappingInfo> getEndpoints(RequestMappingHandlerMapping requestMappingHandlerMapping) {
+
+		if (requestMappingHandlerMapping.getHandlerMethods() != null) {
+			return requestMappingHandlerMapping.getHandlerMethods().keySet();
+		}
+
+		return null;
+	}
+
+}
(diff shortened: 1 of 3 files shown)


marcjansen:
Nice! How well does this play with swagger? Is this related somehow? Is the format you return otherwise standardized?

chrismayer:
This is default format Spring offers us out of the box. I doubt that there is any relation to Swagger or that this is a standardized format.

[springfox](http://springfox.github.io/springfox/) seems to offer a nice way to get a Swagger-compatible API-documentation. Maybe someone finds the time to test and come up with another PR extending this ``EndpointDocController``.

marcjansen:
Ping @buehner wrt to security

I am 👍

buehner:
Nice addition! Thx @chrismayer 

Regarding security we can think about encapsulating the trivial controller logic in a service. The service could easily be secured with an annotation to avoid that everyone can consume this endpoint to retrieve information for possible attacks!?

But i'm not really sure in this point...

chrismayer:
Thanks for the feedback guys!

I agree that we can encapsulate the logic in service, so we are able to secure this interface. 

@buehner: Do you have a suggestion for a security level? I would suggest 

``hasRole(@configHolder.getSuperAdminRoleName()) or hasPermission(returnObject, 'READ')`` 

like is done in other GET requests.

buehner:
I think `hasPermission(returnObject, 'READ')` only makes sense if there is a PermissionEvaluator for the returnObject. But even without a permission evaluator this would not break anything, so :+1: as we can still add a permission evaluator for such a scenario in follow up PRs

chrismayer:
Right @buehner, ``hasPermission(returnObject, 'READ')`` makes no sense here. I just added another commit introducing a service layer, which is secured, so only "admins" can access the endpoint. 

Would be cool if you could have another look at it.

src/shogun2-core/src/main/java/de/terrestris/shogun2/service/EndpointDocService.java
```diff
@@ -0,0 +1,38 @@
...
+import org.springframework.web.servlet.mvc.method.annotation.RequestMappingHandlerMapping;
+
+import de.terrestris.shogun2.web.EndpointDocController;
+
+/**
+ * Service class for the {@link EndpointDocController}.
+ *
+ * @author Christian Mayer
+ */
+@Service("endpoinDocService")
```
buehner: i think there is a small typo (missing "t") :wink:

src/shogun2-core/src/main/java/de/terrestris/shogun2/web/EndpointDocController.java
```diff
@@ -24,25 +28,37 @@
 	@Autowired
 	private RequestMappingHandlerMapping requestMappingHandlerMapping;
 
+	/**
+	 * The service layer instance
+	 */
+	@Autowired
+	@Qualifier("endpoinDocService")
```
buehner: see above. (missing "t") :wink:

buehner:
Just found a small typo. If this is fixed: Feel free to merge :+1:

chrismayer:
Thanks again for your review @buehner! I adressed your comment and will merge now.

PR #268, opened 2017-08-18
Added a parameter to control request method

hwbllmnn:
Can be used in the frontend to switch between requesting with TILED=true/false.


ahennr:
LGTM :+1:

PR #269, opened 2017-08-31
Assure that read-only methods will not persist

buehner:
Last week, we had a strange bug in a project where database entries disappeared after calling a _read-only_ method. The reason was related to the fact that a persisted entity has been manipulated within this transactional method. Even though no `saveOrUpdate` was called, the manipulated entity has been persisted automatically by the underlying transaction manager.

We could fix this bug by annotating the method with `@Transactional(readOnly = true)`

This PR annotates all _read-only_ transactional service methods within the shogun2 project to assure that this will not happen due to shogun2 code :wink:


annarieger:
LGTM, thanks :pineapple:

chrismayer:
Thanks @buehner!

PR #270, opened 2017-09-04
Introduces "describeModel" interface.

KaiVolland:
This introduces a describe-model interface which describes any model as a json-schema.

Makes use of https://github.com/FasterXML/jackson-module-jsonSchema

Does currently not support custom serializers.

### DIFF
--- a/src/shogun2-core/src/main/java/de/terrestris/shogun2/service/ModelDescriptionService.java
+++ b/src/shogun2-core/src/main/java/de/terrestris/shogun2/service/ModelDescriptionService.java
@@ -1,14 +1,19 @@
 package de.terrestris.shogun2.service;
 
 import java.io.IOException;
+import java.util.List;
 
+import javax.annotation.Resource;
+
+import org.apache.log4j.Logger;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.stereotype.Service;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.module.jsonSchema.JsonSchema;
 import com.fasterxml.jackson.module.jsonSchema.JsonSchemaGenerator;
 
-import de.terrestris.shogun2.model.PersistentObject;
 
 /**
  * Service class to describe classes model.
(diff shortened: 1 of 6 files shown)


src/shogun2-core/src/main/java/de/terrestris/shogun2/service/ModelDescriptionService.java
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Service class to describe classes model.
+ *
+ * terrestris GmbH & Co. KG
+ * @author Kai Volland
+ * @date 04.09.2017
+ *
+ */
+@Service("modelDescriptionService")
+public class ModelDescriptionService<E extends PersistentObject> {
```
buehner: `<E extends PersistentObject>` is not necessary if no instances of `PersistentObject` are managed by this service.

src/shogun2-core/src/main/java/de/terrestris/shogun2/service/ModelDescriptionService.java
```diff
@@ -0,0 +1,35 @@
...
+	
+	public JsonSchema getJsonSchema(Class<?> clazz) throws IOException {
+		ObjectMapper mapper = new ObjectMapper();
```
buehner: There is a project-specific `ObjectMapper` that is available as bean (if archetype has been used) and should be used here:

https://github.com/terrestris/shogun2/blob/master/src/shogun2-webapp-archetype/src/main/resources/archetype-resources/src/main/webapp/WEB-INF/__artifactId__-servlet.xml#L54

src/shogun2-core/src/main/java/de/terrestris/shogun2/web/ModelDescriptionController.java
```diff
@@ -28,39 +27,14 @@
...
-					if (!wasNull) throw new RuntimeException(className + " exists in multiple packages!");
-				} catch (ClassNotFoundException e){
-					//not in this package, try another
-				}
-			}
-
-			JsonSchema json = modelDescriptionService.getJsonSchema(foundClass);
-
+			JsonSchema json = modelDescriptionService.getJsonSchema(className);
 			return ResultSet.success(json);
```
marcjansen: We might want to return a `ResultSet.error()` if the `json` is `null`. you decide.

PR #276, opened 2017-09-20
Fix toString methods

buehner:
This will simplify the implementations of `toString` methods. This makes sense as the use of `ToStringBuilder.reflectionToString(this, ToStringStyle.DEFAULT_STYLE);` may impact the performance (because hibernate will lazily query all related data).


dnlkoch:
LGTM!

PR #277, opened 2017-09-25
Fixtures for usage of SHOGun2 with Oracle databases

ahennr:
This PR contains some additions that are needed to get SHOGun2 running on a Oracle 12c database.
In particular:
* Custom Oracle dialect that registers certain LOB types (used e.g. in `File` or `Plugin`)
* Add a more specific implementation of [PhysicalNamingStrategy interface](https://docs.jboss.org/hibernate/orm/5.1/javadocs/org/hibernate/boot/model/naming/PhysicalNamingStrategy.html)  for oracle databases, that:
    * checks for reserved words (determined from Oracle DB's system dictionary`V$RESERVED_WORDS`)
    * checks that length of column name is lesser than 30 characters (as already implemented in `PhysicalNamingStrategyShogun2 `)

In order to use SHOGun2 using an Oracle instance, you've to include these in the following way (e.g. with column name prefix `pref_`)

`hibernate.properties`:
```
hibernate.dialect=de.terrestris.shogun2.util.Shogun2OracleDialect
....
ddl.colPrefix=pref_
....
```

`hibernate.properties`:
```
hibernate.dialect=de.terrestris.shogun2.util.Shogun2OracleDialect
....
ddl.colPrefix=pref_
....
```

In the database context definition, e.g. `project-context-db.xml`, include `columnNamePrefix` bean and set the naming strategy in your `sessionFactory`:

```
<bean id="columnNamePrefix" class="java.lang.String">
    <constructor-arg value="${ddl.colPrefix}"></constructor-arg>
</bean>
...
<bean id="sessionFactory"  class="org.springframework.orm.hibernate5.LocalSessionFactoryBean">
    ....
    <property name="physicalNamingStrategy">
        <bean class="de.terrestris.shogun2.util.naming.OracleNamingStrategyShogun2" />
    </property>
    ....
</bean>

```

Please review @terrestris/devs

### DIFF
--- a/src/shogun2-core/src/main/java/de/terrestris/shogun2/model/layer/source/TileWmsLayerDataSource.java
+++ b/src/shogun2-core/src/main/java/de/terrestris/shogun2/model/layer/source/TileWmsLayerDataSource.java
@@ -37,8 +37,7 @@ public class TileWmsLayerDataSource extends ImageWmsLayerDataSource {
 	 * Whether to request the layer with TILED=true.
 	 */
 	@Column(name = "REQUEST_WITH_TILED")
-	@ColumnDefault(value = "true")
-	private Boolean requestWithTiled = true;
+	private Boolean requestWithTiled = Boolean.TRUE;
 
 	/**
 	 * default constructor
--- a/src/shogun2-core/src/main/java/de/terrestris/shogun2/util/naming/PhysicalNamingStrategyShogun2.java
+++ b/src/shogun2-core/src/main/java/de/terrestris/shogun2/util/naming/PhysicalNamingStrategyShogun2.java
@@ -112,7 +112,7 @@ public class PhysicalNamingStrategyShogun2 implements PhysicalNamingStrategy, Se
 			// identifier limit of 30 chars -->
 			// http://stackoverflow.com/a/756569
 			return LENGTH_LIMIT_ORACLE;
-		} if (context.getDialect() instanceof Shogun2OracleDialect) {
+		} else if (context.getDialect() instanceof Shogun2OracleDialect) {
 			// identifier limit of 30 chars -->
 			return LENGTH_LIMIT_ORACLE;
 		} else if (dialectName.startsWith("PostgreSQL")) {
--- a/src/shogun2-core/src/main/java/de/terrestris/shogun2/util/naming/PhysicalNamingStrategyShogun2.java
+++ b/src/shogun2-core/src/main/java/de/terrestris/shogun2/util/naming/PhysicalNamingStrategyShogun2.java
@@ -1,11 +1,14 @@
 package de.terrestris.shogun2.util.naming;
 
+import de.terrestris.shogun2.util.dialect.Shogun2OracleDialect;
 import org.apache.commons.lang3.StringUtils;
 import org.hibernate.boot.model.naming.Identifier;
-import org.hibernate.boot.model.naming.PhysicalNamingStrategyStandardImpl;
+import org.hibernate.boot.model.naming.PhysicalNamingStrategy;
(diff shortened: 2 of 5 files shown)


src/shogun2-core/src/main/java/de/terrestris/shogun2/model/layer/source/TileWmsLayerDataSource.java
```diff
@@ -37,8 +37,7 @@
 	 * Whether to request the layer with TILED=true.
 	 */
 	@Column(name = "REQUEST_WITH_TILED")
-	@ColumnDefault(value = "true")
-	private Boolean requestWithTiled = true;
+	private Boolean requestWithTiled = Boolean.TRUE;
```
buehner: Is there a reason for this change? I think this will lead to different behaviour:

Using `@ColumnDefault` will apply the `DEFAULT` value to the associated column in the DDL, i.e. if you use the hibernate ddl UPDATE option on a database that is missing this column, the values in the database will initially be "true" afterwards.

With your change, these values would be null in case of UPDATE, right?

src/shogun2-core/src/main/java/de/terrestris/shogun2/util/naming/PhysicalNamingStrategyShogun2.java
```diff
@@ -94,16 +104,17 @@ private Identifier convertToLimitedLowerCase(JdbcEnvironment context, Identifier
...
+	protected Integer getIdentifierLengthLimit(JdbcEnvironment context) {
 		// https://docs.jboss.org/hibernate/orm/5.0/javadocs/org/hibernate/dialect/package-summary.html
 		String dialectName = context.getDialect().getClass().getSimpleName();
 
 		if (dialectName.startsWith("Oracle")) {
 			// identifier limit of 30 chars -->
 			// http://stackoverflow.com/a/756569
 			return LENGTH_LIMIT_ORACLE;
-
+		} if (context.getDialect() instanceof Shogun2OracleDialect) {
```
buehner: I think this should be `else if` instead of `if`, right?

src/shogun2-core/src/main/java/de/terrestris/shogun2/model/layer/source/TileWmsLayerDataSource.java
ahennr: The `value` provided in `@ColumnDefault` annotation as column default is vendor specific since it will be applied to DDL. In case of Oracle databases it has to be a numerical value (1 in this case), here. Using `"true"` loads to the following (errornous) DDL:

```
create table tilewmslayerdatasources (request_with_tiled number(1,0) default true, 
    id number(10,0) not null, tilegrid_id number(10,0), primary key (id))
```

If we use a non-nullable column here (e.g. `@Column(name = "REQUEST_WITH_TILED", nullable = false)` ) these values would not be null on hibernate update, right?

src/shogun2-core/src/main/java/de/terrestris/shogun2/util/naming/PhysicalNamingStrategyShogun2.java
ahennr: Nice catch! Correct, fixed in f162d9ab49

src/shogun2-core/src/main/java/de/terrestris/shogun2/model/layer/source/TileWmsLayerDataSource.java
buehner: Ok. I understand. I don't think that this is a major issue, so in my eyes you can leave it as proposed

buehner:
Feel free to merge

ahennr:
THX for review @buehner and @marcjansen

PR #278, opened 2017-09-26
Added simple jpa jsonb converter

hwbllmnn:

### DIFF
--- a/src/shogun2-core/src/main/java/de/terrestris/shogun2/hibernate/SimpleJsonbConverter.java
+++ b/src/shogun2-core/src/main/java/de/terrestris/shogun2/hibernate/SimpleJsonbConverter.java
@@ -1,0 +1,51 @@
+package de.terrestris.shogun2.hibernate;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.apache.log4j.Logger;
+
+import javax.persistence.AttributeConverter;
+import javax.persistence.Converter;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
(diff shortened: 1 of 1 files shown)


src/shogun2-core/src/main/java/de/terrestris/shogun2/hibernate/SimpleJsonbConverter.java
```diff
@@ -0,0 +1,51 @@
...
+ * btree indexes.
+ */
+@Converter(autoApply = true)
+public class SimpleJsonbConverter implements AttributeConverter<Map<String, String>, String> {
+
+    private static final Logger LOG = Logger.getLogger(SimpleJsonbConverter.class);
+
+    @Override
+    public String convertToDatabaseColumn(Map<String, String> attribute) {
+        ObjectMapper mapper = new ObjectMapper();
```
buehner: I don't think that this will make a difference, but maybe we should use an instance of `Shogun2JsonObjectMapper` here?

https://github.com/terrestris/shogun2/blob/master/src/shogun2-core/src/main/java/de/terrestris/shogun2/util/json/Shogun2JsonObjectMapper.java

hwbllmnn: I wasn't aware of this one, thanks! I think it might still make a difference, I'm sure we'll be storing timestamps in there, so I changed it.

marcjansen:
Will this enable user to store JSON in the appropriate columb type in postgres?

hwbllmnn:
This can be used to map jsonb fields with 'simple' content (objects with a simple key/value mapping like {"name": "peter", "somekey": 123}) with hibernate.

buehner:
:+1: from me. feel free to merge

PR #280, opened 2017-10-12
Interceptor wms reflector option for interceptor

ahennr:
This PR introduces an option for GeoServer interceptor (parameter `useReflect`) to call [WMS reflector endpoint](http://docs.geoserver.org/latest/en/user/tutorials/wmsreflector.html) of GeoServer.

An example call to interceptor (layer `dumpy` in namespace `humpty`) would be:

```
http://localhost:9090/geoserver.action?layers=humpty:dumpy&useReflect=true
```


buehner:
LGTM :+1:

ahennr:
THX for review @buehner

marcjansen:
I think this breaks backwards compatibility by effectively removing the previously `public` method `URI getGeoServerBaseURIFromNameSpace(String geoServerNamespace)`.

I think this should at least deserved a release.

PR #281, opened 2017-10-16
Allow interception of W3DS requests

buehner:
This enables the interceptor to forward W3DS requests. Tests have also been added.


buehner:
Since W3DS has not become an official OGC standard, i'll close this again and open another PR soon...

buehner:
I tried to find another approach to support W3DS but this was not as easy as expected because a lot of code is based on enums. This leads to different problems:

* enums are not extendible (e.g. for a project specific solution)
* In the `InterceptorRule` model, the `service` enum is configured with `@Column(nullable = false)`. I tried to introduce some kind of unspecific/generic rule (e.g. without a service information), but changing the column annotation to `nullable = true` would break the backwards-compatibility

So i'll re-open this PR again. Introducing W3DS here should not hurt anybody...

weskamm:
Looks ok to me so far, maybe we need to replace the enums in a future PR...
Merge at will!

PR #282, opened 2017-10-17
Reduce the interceptor logging on DEBUG

buehner:
In my eyes the geoserver interceptor was logging too much on the DEBUG level, which lead to messy logs during development (and may even slow down the performance) as a lot of "trace-level" information like HTTP headers etc were logged.

This will PR changes some logs from DEBUG to TRACE, but for an intercepted request, the following logs will still remain on DEBUG:

```
2017-10-17 14:57:27,995 DEBUG [de.terrestris.shogun2.service.GeoServerInterceptorService] - Finding the GeoServer base URI by the provided EndPoint: foo:bar
2017-10-17 14:57:27,999 DEBUG [de.terrestris.shogun2.service.GeoServerInterceptorService] - The corresponding GeoServer base URI is: http://localhost:8080/geoserver/foo/ows
2017-10-17 14:57:27,999 DEBUG [de.terrestris.shogun2.util.interceptor.OgcMessageDistributor] - Request is ALLOWED, not intercepting the request.
2017-10-17 14:57:28,184 DEBUG [de.terrestris.shogun2.util.interceptor.OgcMessageDistributor] - Response is ALLOWED, not intercepting the response.
```

### DIFF
--- a/src/shogun2-core/src/main/java/de/terrestris/shogun2/service/GeoServerInterceptorService.java
+++ b/src/shogun2-core/src/main/java/de/terrestris/shogun2/service/GeoServerInterceptorService.java
@@ -154,7 +154,7 @@ public class GeoServerInterceptorService {
 	private OgcMessage getOgcMessage(MutableHttpServletRequest mutableRequest)
 			throws InterceptorException, IOException {
 
-		LOG.debug("Building the OGC message from the given request.");
+		LOG.trace("Building the OGC message from the given request.");
 
 		OgcMessage ogcMessage = new OgcMessage();
 
@@ -217,7 +217,7 @@ public class GeoServerInterceptorService {
 			LOG.debug("No interceptor rule found for the response.");
 		}
 
-		LOG.debug("Successfully build the OGC message: " + ogcMessage);
+		LOG.trace("Successfully build the OGC message: " + ogcMessage);
 
 		return ogcMessage;
 	}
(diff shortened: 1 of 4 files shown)


src/shogun2-core/src/main/java/de/terrestris/shogun2/service/GeoServerInterceptorService.java
```diff
@@ -320,8 +320,8 @@ public boolean evaluate(InterceptorRule rule) {
 					"least the basic sets of rules (e.g. ALLOW all WMS " +
 					"requests) when using the interceptor.");
 			throw new InterceptorException("No interceptor rule found.");
-		} else {
-			LOG.debug("Identified the following rule as most the specific " +
+		} else if(LOG.isTraceEnabled()) {
+			LOG.trace("Identified the following rule as most the specific " +
 					"one: " + mostSpecificRule);
```
marcjansen: I thi the toString method of the rules will result in added newlines here. while we are at it, I'd like to have only one line of logging here.

This is non blocking of course.

marcjansen:
LGTM. Please merge at will.

src/shogun2-core/src/main/java/de/terrestris/shogun2/service/GeoServerInterceptorService.java
buehner: Yes. Maybe newlines would be added here, but this depends on the `toString` implementation of the `InterceptorRule` model, right?

The if-condition to check whether trace is enabled has been added to avoid that some kind of lazy loading (of the rule) would be triggered, but i think that this will not happen as the `InterceptorRule` model does not have any dependent sub-entities (but only primitive fields).

PR #283, opened 2017-11-08
Make getGeoServerBaseURIFromNameSpace method be public again

annarieger:
Visibility of the method `getGeoServerBaseUriFromNameSpace` was [decreased](https://github.com/terrestris/shogun2/pull/280/files#diff-3ca41da6aab8eea4ad1630eec73b2825L374) from `public` to `private` in #280. That has broken backwards compatibility for some projects which make use of this method.

This PR makes `getGeoServerBaseURIFromNameSpace` be public again.

Please review @buehner @ahennr


buehner:
:+1: I'll merge now

PR #287, opened 2017-12-13
Introduce HTTP-forward-proxy

ahennr:
This PR introduces a simple HTTP forward proxy which forwards HTTP requests (with params) to given `baseUrl`. In addition:
* Check if requested host is contained in whitelist (comma separated list having property key `proxy.whitelist`)
* Supports forwarding of HTTP GET, POST and FormMultipartPost
* Currently, response headers if proxied request are not modified but simply passed to response 
* Adds log4j properties file to configure logging for tests

Plz review.


src/shogun2-core/src/test/resources/log4j.properties
```diff
@@ -0,0 +1,66 @@
+#################################################
+# LOGGING PROPERTIES FOR TESTING PURPOSE
```
buehner: As we have migrated to log4j v2 with one of the latest merges, we should replace this config with an appropriate log4j v2 config xml. Just have a look here:

* https://github.com/terrestris/shogun2/blob/master/src/shogun2-core/src/test/resources/log4j2.xml
* https://github.com/terrestris/shogun2/blob/master/src/shogun2-webapp-archetype/src/main/resources/archetype-resources/src/main/resources/log4j2.xml

ahennr: Since @hwbllmnn already introduced a corresponding log4j2 config in the test resources, I've simply removed `log4j.properties` in 42ff914

ahennr:
THX for review :top:

PR #289, opened 2017-12-19
Add hibernate caching, bump some dependency versions

hwbllmnn:
This updates some dependencies to recent versions, most notably log4j2 (log4j 1 is at EOL since ~2015).

Most notably, this adds hibernate cache annotations and also uses JOINs where possible. Also, GenericHibernateDao#createDistinctRootEntityCriteria uses setCacheable with the value of the hibernate.cache.use_query_cache property, effectively caching everything if set to true.

@terrestris/devs @buehner please review carefully

We probably should make a 0.1.4 release before merging this, as projects probably need some updates when using this version.

Alternatively we could create a 0.2.x release branch and merge this there.


marcjansen:
I like this, very nice work, thanks.

`+1` for a release before merging this.

Can you update the archetype so we can see which changes would be needed there?

Do people need to change the log4j configurations they have?

Related: Is there any chance that we can control the `@Fetch` configuration at build time of the actual application, not at build time of this framework? I find it odd that we decide which is the best way in the framework, and I faintly remeber switching between configurations often. This last comment should not block this from being merged.

I think this deserves more review work from others.

buehner:
I just had a very quick look at this for now and this PR looks really great! I'll have a deeper look at this later.

hwbllmnn:
Regarding the `@Fetch` configuration, you can override this manually when querying (eg. `criteria.setFetchMode(path, FetchMode.SUBSELECT);`).

The log4j configurations need to be adapted, unfortunately, but I provided a template in `src/test/resources/log4j2.xml` which is pretty straightforward IMHO.

Regarding the archetype, I'll have a look into it and report back.

hwbllmnn:
Archetype changes have been added in e2c27d9 (example log4j2.xml, hibernate.properties, default cache config).

src/shogun2-core/src/test/resources/META-INF/hibernate.properties
```diff
@@ -5,7 +5,7 @@ hibernate.id.new_generator_mappings=true
 hibernate.hbm2ddl.auto=create
 
 # caching
-#hibernate.cache.use_query_cache=true
+hibernate.cache.use_query_cache=false
```
weskamm: did you experience issues having set this to true? if so, which ones?

weskamm:
i have left just a minor question, looking good to me.

src/pom.xml
```diff
@@ -135,7 +135,7 @@
         <spring.version>4.3.11.RELEASE</spring.version>
         <spring-security.version>4.2.3.RELEASE</spring-security.version>
         <!-- TODO: Raise log4j to version 2 -->
```
ahennr: Very, very minor: This TODO is done by your change, isn't it? So it can be removed.

src/shogun2-core/src/test/resources/META-INF/hibernate.properties
hwbllmnn: Not really, but since this was implicitly the default before, I thought I'd leave it at that. I just need the property to be set, since I'm referencing it in the code.

src/pom.xml
```diff
@@ -135,7 +135,7 @@
...
-        <log4j.version>1.2.17</log4j.version>
+        <log4j.version>2.10.0</log4j.version>
```
ahennr: Did you add an log4j config file for the tests as well?

src/pom.xml
hwbllmnn: True, removed it.

src/pom.xml
hwbllmnn: No, for the tests I just stuck with the `slf4j-simple` logger. But I converted the `log4j.properties` example in the archetype to a new `log4j2.xml`.

weskamm:
I see a lot of deprecation warnings using this code in another project:

WARN [org.hibernate.orm.deprecation] - HHH90000022: Hibernate's legacy org.hibernate.Criteria API is deprecated; use the JPA javax.persistence.criteria.CriteriaQuery instead

I think this is related, could you please check this?

hwbllmnn:
@weskamm The warnings you're seeing are related to the projects' switch to Hibernate 5, where said API has become deprecated. I agree we should update SHOGun to use the JPA criteria API, but that's unrelated to this PR and also a lot of work 😿

hwbllmnn:
I've just added a commit with a few changes to the archetype so the generated webapp will start again with jetty. Changes include:

* no more JDK 7 support (log4j2 uses multi release jars: https://stackoverflow.com/questions/46878035/error-scanning-entry-meta-inf-versions-9-org-apache-logging-log4j-util-stackloca)
* changed scope of servlet api jars to `provided` so they won't cause failures on startup
* removed log4j 1 specific stuff from web.xml
* added missing hibernate cache config options

buehner:
@terrestris/devs

the release v0.1.4 is done, so this could be merged now. But i think we still need to remove `openjdk7` from the `.travis.yml` as we won't support Java 7 anymore after merging this. Correct?

ahennr:
Absolutely, Java7 dependency needs to be removed.
Currently we only use OracleJDK8 in our travis builds. We should include OpenJDK8 as well, shouldn't we?

src/shogun2-core/src/test/resources/META-INF/hibernate.properties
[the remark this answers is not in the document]
weskamm: See https://github.com/terrestris/shogun2/pull/289#pullrequestreview-84695476

PR #291, opened 2018-01-22
Add boundless maven repo

buehner:
As http://download.osgeo.org/webdav/geotools is not responding reliable, we may have better luck by using https://repo.boundlessgeo.com/main/

### DIFF
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -72,6 +72,10 @@
             </releases>
         </repository>
         <repository>
+            <id>OSGeo GeoTools repo</id>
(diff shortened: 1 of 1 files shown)


buehner:
It seems that there is a dependency on `jgridshift`, which is only available in the osgeo-geotools repo (which seems to be available again), so i added the osgeo repo again. (But adding the boundless repo does not hurt regarding redundancy).

buehner:
As this will only add an additional repo to enhance redundancy, but change nothing else, i'll merge now...

PR #292, opened 2018-01-23
Avoid exception when running test

buehner:
During the tests an exception occured (but the tests did not fail). By ignoring the `javax.management.*` classes, the exceptions will not occur anymore...


hwbllmnn:
Nice, merging now.

PR #295, opened 2018-01-26
Fix joins

hwbllmnn:
This fixes the hashcode null pointers when logging in and using the REST API requesting modules.


buehner:
:+1: very important fix! thank you. i'll merge now

PR #297, opened 2018-01-26
Added plugin management, added more dependencies to mgmt

hwbllmnn:
This will fix dependency convergence issues for projects using the SHOGun2 parent pom as parent. Also fixes/updates plugin dependencies.

### DIFF
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -150,6 +150,8 @@
         <hikari.version>2.7.6</hikari.version>
         <h2.version>1.4.196</h2.version>
 
+        <jdom.version>1.1.3</jdom.version>
+
         <!-- Testing -->
         <junit.version>4.12</junit.version>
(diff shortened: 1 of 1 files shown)


src/pom.xml
```diff
@@ -321,6 +331,55 @@
     </profiles>
 
     <build>
+        <pluginManagement>
+            <plugins>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-clean-plugin</artifactId>
+                    <version>3.0.0</version>
```
buehner: Could all the versions in the `pluginManagement` section also be extracted to properties?

hwbllmnn: Done in 2dd454f.

buehner:
Thanks a lot, i'll merge now

PR #299, opened 2018-02-05
Fix several minor issues in the HttpProxyService

marcjansen:
Title says it all. A follow up of #287.


marcjansen:
Thanks for the review.

PR #304, opened 2018-03-05
MutableHttpServletRequest - Adds "setter" for internal `cachedInputStream` and corresponding tests

ahennr:
Title says it all.


src/shogun2-core/src/main/java/de/terrestris/shogun2/util/interceptor/MutableHttpServletRequest.java
```diff
@@ -159,15 +163,15 @@ public static String getRequestParameterValue(HttpServletRequest httpServletRequ
 
 	/**
 	 *
-	 * @param url
+	 * @param url The URI to as instance of {@link String}
```
dnlkoch: Missing `(…) set (…)`.

ahennr:
THX for review @dnlkoch .

PR #305, opened 2018-03-07
Add ldap service

hwbllmnn:
Adds a simple ldap service that can authenticate users against an ldap. It can also extract roles/groups from ldap.

### DIFF
--- a/src/shogun2-core/src/main/java/de/terrestris/shogun2/service/LdapService.java
+++ b/src/shogun2-core/src/main/java/de/terrestris/shogun2/service/LdapService.java
@@ -1,0 +1,67 @@
+package de.terrestris.shogun2.service;
+
+import org.apache.log4j.Logger;
+import org.springframework.ldap.core.AttributesMapper;
+import org.springframework.ldap.core.LdapTemplate;
+import org.springframework.stereotype.Service;
+
+import javax.naming.NamingEnumeration;
+import javax.naming.NamingException;
+import javax.naming.directory.Attributes;
+import java.util.ArrayList;
+import java.util.List;
+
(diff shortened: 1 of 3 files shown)


src/shogun2-core/src/main/java/de/terrestris/shogun2/service/LdapService.java
```diff
@@ -0,0 +1,67 @@
...
+     */
+    public List<String> getGroups(String username, String property) {
+        final List<String> result = new ArrayList<>();
+        ldapTemplate.search(query().where("cn").is(username), new AttributesMapper<String>() {
+            public String mapFromAttributes(Attributes attrs) throws NamingException {
+                NamingEnumeration<?> ous = attrs.get(property).getAll();
+                while(ous.hasMore()) {
+                    result.add((String)ous.next());
+                }
+                return "";
```
weskamm: hmm??

hwbllmnn: I added a comment explaining why an empty string is returned here.

src/shogun2-core/src/main/java/de/terrestris/shogun2/service/LdapService.java
```diff
@@ -0,0 +1,70 @@
...
+    /**
+     * Set the ldap template property for ldap access.
+     * @param ldapTemplate the template to set
+     */
+    public void setLdapTemplate(LdapTemplate ldapTemplate) {
+        this.ldapTemplate = ldapTemplate;
+    }
+
+    /**
+     *Authenticate against ldap.
```
marcjansen: Missing a space.

PR #308, opened 2018-03-26
Make usage of whitelist in HTTPProxy optional

weskamm:
Makes the usage of the whitelist in HTTPProxy optional.
By default, it will still remain active.
This is useful when e.g. using the proxy for GetCapabilities requests, as these URLs cannot be added to the whitelist.


weskamm:
Thanks, will merge now...

PR #310, opened 2018-03-27
Enhance models of WMTS layers

ahennr:
This PR adds a bunch of further attributes to the existing `WmtsLayerDataSource` and adds a `WmtsTileGrid` model which can be used w.g. with a [WMTS source](https://openlayers.org/en/latest/apidoc/ol.source.WMTS.html) of OpenLayers .

### DIFF
--- a/src/shogun2-core/src/main/java/de/terrestris/shogun2/model/layer/source/WmtsLayerDataSource.java
+++ b/src/shogun2-core/src/main/java/de/terrestris/shogun2/model/layer/source/WmtsLayerDataSource.java
@@ -1,13 +1,15 @@
 package de.terrestris.shogun2.model.layer.source;
 
-import javax.persistence.Cacheable;
-import javax.persistence.Entity;
-import javax.persistence.Table;
+import javax.persistence.*;
 
+import de.terrestris.shogun2.model.layer.util.TileGrid;
+import de.terrestris.shogun2.model.layer.util.WmtsTileGrid;
 import org.apache.commons.lang3.builder.EqualsBuilder;
 import org.apache.commons.lang3.builder.HashCodeBuilder;
 import org.hibernate.annotations.Cache;
 import org.hibernate.annotations.CacheConcurrencyStrategy;
+
+import java.util.List;
 
 /**
  * Class representing a layer source for tile data from WMTS servers.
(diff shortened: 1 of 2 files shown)


src/shogun2-core/src/main/java/de/terrestris/shogun2/model/layer/util/WmtsTileGrid.java
```diff
@@ -0,0 +1,67 @@
...
+ * Class representing a WMTS tile grid
+ *
+ * @author Andre Henn
+ */
+@Entity
+@Table
+public class WmtsTileGrid extends TileGrid {
+
+	// String array holding the matrix IDs.
+	private String[] matrixIds;
```
buehner: How would this be persisted?

src/shogun2-core/src/main/java/de/terrestris/shogun2/model/layer/source/WmtsLayerDataSource.java
```diff
@@ -63,6 +239,14 @@ public boolean equals(Object obj) {
...
         return new EqualsBuilder().
             appendSuper(super.equals(other)).
+            append(getFormat(), other.getFormat()).
+            append(getRequestEncoding(), other.getRequestEncoding()).
+            append(getWmtsLayer(), other.getWmtsLayer()).
+            append(getMatrixSet(), other.getMatrixSet()).
+            append(getProjection(), other.getProjection()).
+            append(getWmtsStyle(), other.getWmtsStyle()).
+            append(getTileGrid(), other.getTileGrid()).
+            append(getUrls(), other.getUrls()).
```
buehner: Maybe we should not use getTileGrid and getUrls here as this may lead to "expensive" DB queries?

src/shogun2-core/src/main/java/de/terrestris/shogun2/model/layer/source/WmtsLayerDataSource.java
```diff
@@ -45,6 +213,14 @@ public int hashCode() {
...
         return new HashCodeBuilder(11, 19).
             appendSuper(super.hashCode()).
+            append(getFormat()).
+            append(getRequestEncoding()).
+            append(getWmtsLayer()).
+            append(getMatrixSet()).
+            append(getProjection()).
+            append(getWmtsStyle()).
+            append(getTileGrid()).
+            append(getUrls()).
```
buehner: Maybe we should not use getTileGrid and getUrls here as this may lead to "expensive" DB queries?

src/shogun2-core/src/main/java/de/terrestris/shogun2/model/layer/util/WmtsTileGrid.java
ahennr: In order to be consistent with the other models in SHOGun2, I'll switched to `List<String>` (annotated with `@ElementCollection(targetClass = String.class)`) here.

src/shogun2-core/src/main/java/de/terrestris/shogun2/model/layer/source/WmtsLayerDataSource.java
ahennr: As collections are currently included in almost every model of SHOGun2, I've appended `getTileGrid` and `getUrls` to `equals` and `hashCode` method.

However, IMHO,  we could think about removing these in the models since all models inherit from `PersistentObject` which generates a unique id that can be used for equality check, for example.

buehner: I don't think that this will make a big difference, so feel free to merge ;-)
I don't think that using the id field from `PersistenObject` is a good choice since it is possible that a non-persisted entity is already managed in a set and will be persisted afterwards, which may lead to unexpected problems (at least in theory :wink: )

buehner:
:+1: feel free to merge

ahennr:
THX for review @buehner

PR #312, opened 2018-04-09
Fix determination of most specific rule

hwbllmnn:
Previously not all rules were considered, the first matching one would win.


hwbllmnn:
@ahennr Actually the tests already check for this behaviour, it was just sheer luck that they would pass previously. I've changed the order of the test rules so they would fail with the previous code.

hwbllmnn:
1.0.2 and 0.1.7 have been released and should show up on central in a couple of hours.

marcjansen:
Very nice. Thanks.

PR #316, opened 2018-04-30
Adjust column content length

dnlkoch:
This increases the maximum content length of the fields `layerNames` and `layerStyles` from 255 to 2048 characters.


buehner:
Will this change also apply for existing projects (e.g. when booting with hibernate update mode)?

Otherwise: LGTM :+1:

buehner:
According to this post, it seems that the defintion of existing columns will not be changed when booting in update mode: https://stackoverflow.com/a/26681592

So we should keep this in mind! Feel free to merge @dnlkoch

PR #318, opened 2018-05-15
Fix snapshot releases

buehner:
This hopefully leads to the snapshot releases @hwbllmnn ;-)

### DIFF
--- a/scripts/snapshot-release.sh
+++ b/scripts/snapshot-release.sh
@@ -3,7 +3,7 @@
 # stop at first command failure
 set -e
 
-if [ "$TRAVIS" != "true" ]; then
+if [ $TRAVIS != "true" ]; then
     echo "This script is supposed to be run inside the travis environment."
     exit 1
 fi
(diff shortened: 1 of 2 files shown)


hwbllmnn:
Nice catch ;-)

scripts/snapshot-release.sh
```diff
@@ -3,7 +3,7 @@
 # stop at first command failure
 set -e
 
-if [ "$TRAVIS" != "true" ]; then
+if [ $TRAVIS != "true" ]; then
```
marcjansen: This was not related, was it? if the variable evaluates to the empty string, this will result in a script error, which is hard to debug. The quotes were a safe-guard, because comparing strings in bash can be hard.

Same comment applies below.

hwbllmnn: Actually, the quotes were added automatically by the travis command line tool to add encrypted passwords. Since the tool also broke yaml indent we just reverted to the original state (with just the added secrets).

marcjansen: anyhow this is not a big problem. it should be fixed nonetheless.

buehner: @marcjansen I think this was my fault. I didnt know that the quotes were a safe-guard.
@hwbllmnn I think you're mixing this up with the travis yaml. I dont think that the travis command line tool changed the bash script!?

I created a PR to fix this: https://github.com/terrestris/shogun2/pull/319

hwbllmnn: True, I mixed it up :roll_eyes:

PR #321, opened 2018-05-24
Add additivity parameter to avoid duplicate logs

hwbllmnn:
@terrestris/devs FYI, this is why we had duplicate logs since moving to log4j2.


buehner:
Thank you @hwbllmnn 
I will merge now!

PR #324, opened 2018-06-22
Allow forwarding of request headers in interceptor

buehner:
Allow forwarding of whitelisted http request headers in the interceptor

Please review @terrestris/devs


hwbllmnn:
Nice one, please merge at will!

PR #325, opened 2018-06-23
WIP: Add experimental json blob storage with automatic conversion

hwbllmnn:
Not really meant for merging, this is just a showcase how a json blob based storage & migration could work.


buehner:
I'll close this now as it was just a proof of concept

PR #331, opened 2018-12-13
Migrate to latest JTS

buehner:
This will migrate to the latest JTS, which is used by the latest versions of geotools. This is necessary to remain compatible to the latest geoserver versions. Major change in JTS is that the package name has changed.

Related links:

* https://github.com/locationtech/jts/blob/master/MIGRATION.md
* https://github.com/bedatadriven/jackson-datatype-jts/pull/18#issuecomment-399002136

### DIFF
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -62,7 +62,6 @@
         </repository>
     </repositories>
 
-
     <pluginRepositories>
         <pluginRepository>
             <id>nexus.terrestris.de-plugins</id>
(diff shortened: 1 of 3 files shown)


marcjansen:
Does this mean that we can no longer talk to older GS versions?

buehner:
Yes. As soon as this is merged (or a new version is released), we are only compatible to JTS >= v1.15

But as the latest versions of geotools/geoserver are using the latest JTS, we should also do this step.

hwbllmnn:
LGTM. Maybe we should release this as 1.1.0 then?

buehner:
yes. as soon as this is merged, i'll update the release notes and then a v1.1.0  release makes sense

PR #336, opened 2018-12-19
Fixes hashcode and equals implementation

### LINKED ISSUE #290: Problem with equals implementation
The `PersistentObject` implements `#equals`/`#hashCode` by comparing the modified and created timestamps. This causes problems in certain situations, especially when testing and using mocked up entities. On slow systems, two entities that should have been considered equal will not be equal due to differing timestamps.

We should think about whether to include said fields in equality tests.

hwbllmnn:
Fixes #290 

@buehner Please review.

### DIFF
--- a/src/shogun2-core/src/main/java/de/terrestris/shogun2/model/PersistentObject.java
+++ b/src/shogun2-core/src/main/java/de/terrestris/shogun2/model/PersistentObject.java
@@ -4,34 +4,20 @@
  */
 package de.terrestris.shogun2.model;
 
-import java.io.Serializable;
-import java.util.HashMap;
-import java.util.Map;
-
-import javax.persistence.Cacheable;
-import javax.persistence.Column;
-import javax.persistence.GeneratedValue;
-import javax.persistence.GenerationType;
-import javax.persistence.Id;
-import javax.persistence.JoinColumn;
-import javax.persistence.JoinTable;
-import javax.persistence.ManyToMany;
-import javax.persistence.MappedSuperclass;
-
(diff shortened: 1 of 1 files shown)


buehner:
I am not sure whether it is a good idea to use the ID field for hashCode and equals implementation. I have in mind that the ID was not involved for reasons!

In the persistence context (Hibernate/JPA) this may be problematic in some scenarios:
https://vladmihalcea.com/how-to-implement-equals-and-hashcode-using-the-jpa-entity-identifier/

*When the entity was first stored in the Set, the identifier was null. After the entity was persisted, the identifier was assigned to a value that was automatically generated, hence the hashCode differs. For this reason, the entity cannot be found in the Set after it got persisted.*
...
*The entity identifier can be used for equals and hashCode, but only if the hashCode returns the same value all the time.*

So maybe we should only rely on the getClass and created values?

buehner:
Here is more stuff: https://stackoverflow.com/questions/5031614/the-jpa-hashcode-equals-dilemma

hwbllmnn:
That's a good point @buehner. I've removed the id, so only the class remains. I think this is still better than the created/modified comparison. What do you think?

buehner:
But wouldn't this mean that two instances of the same class could have the same hashcode, which is not intended!?

As the created field is set in the constructor (`this.created = DateTime.now();`), in my eyes a combination of created (which will not change afterwards) and the class makes more sense!?

hwbllmnn:
I'm not sure how much sense an equality definition on `PersistedObject` makes anyway, but you're right, it doesn't hurt to add the `created` field; done.

PR #338, opened 2019-01-02
Forward GWC headers for missed caches

ahennr:
Title says it all. In particular, `geowebcache-miss-reason` will be forwarded


buehner:
:+1:  merge plz

PR #339, opened 2019-01-28
Adds a jsonb user type to handle more complex documents

hwbllmnn:
This enables us to incorporate PostgreSQL jsonb fields into object graphs.

Use it like this:

```java
    @org.hibernate.annotations.Type(type="de.terrestris.shogun2.hibernate.JsonbUserType")
    private Map<String, Object> custom;
```

and add the appropriate getter/setter as needed. The stored json can be of any depth and will be incorporated as the value of the `custom` field of the annotated entity transparently.

### DIFF
--- a/src/shogun2-core/src/main/java/de/terrestris/shogun2/hibernate/JsonbUserType.java
+++ b/src/shogun2-core/src/main/java/de/terrestris/shogun2/hibernate/JsonbUserType.java
@@ -1,29 +1,24 @@
 package de.terrestris.shogun2.hibernate;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import de.terrestris.shogun2.util.json.Shogun2JsonObjectMapper;
 import org.hibernate.HibernateException;
 import org.hibernate.engine.spi.SharedSessionContractImplementor;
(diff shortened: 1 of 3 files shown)


src/shogun2-core/src/main/java/de/terrestris/shogun2/hibernate/JsonbUserType.java
```diff
@@ -0,0 +1,123 @@
...
+    public Object nullSafeGet(ResultSet rs, String[] names, SharedSessionContractImplementor session, Object owner) throws HibernateException, SQLException {
+        if (!rs.wasNull()) {
+            Shogun2JsonObjectMapper mapper = new Shogun2JsonObjectMapper();
+            try {
+                String string = rs.getString(names[0]);
+                if (string == null) {
+                    return null;
+                }
+                HashMap map = mapper.readValue(string, HashMap.class);
+                System.out.println(map);
```
buehner: :1st_place_medal:

hwbllmnn: :duck: https://github.com/terrestris/shogun-core/pull/341

PR #344, opened 2019-03-27
Add support for custom endpoints

hwbllmnn:
This supports interceptor urls like `/geoserver.action/somename`. If the layer sources are configured accordingly and the new standard WMS interceptors are configured, the interceptor will only support the layers that are configured with this path at the custom endpoint and filter out the rest. `GetCapabilities`, `GetMap`, `GetFeatureInfo` and `GetLegendGraphic` are supported.

### DIFF
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -4,7 +4,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>de.terrestris</groupId>
     <artifactId>shogun2</artifactId>
-    <version>1.0.5</version>
+    <version>1.0.6-SNAPSHOT</version>
     <packaging>pom</packaging>
     <name>SHOGun2</name>
     <description>SHOGun2 is a Java based WebGIS framework</description>
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -50,6 +50,11 @@
 
     <repositories>
         <repository>
+            <id>terrestris</id>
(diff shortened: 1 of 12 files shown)


buehner:
Is this a new feature for the 1.0.x-master branch only? Or shouldnt this also be merged to master?

hwbllmnn:
> Is this a new feature for the 1.0.x-master branch only? Or shouldnt this also be merged to master?

PR for the master is coming up, I just need this for a 1.0.x project as well.

src/pom.xml
```diff
@@ -4,7 +4,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>de.terrestris</groupId>
     <artifactId>shogun2</artifactId>
-    <version>1.0.5</version>
```
buehner: Is this a problem? It seems that there are two commits missing (after the last release to set the new snapshot version)

hwbllmnn: Well, I don't know where those commits are. We didn't have a `1.0.x` branch yet, so I created one from the `1.0.5` tag, hence the custom version update.

hwbllmnn:
FYI, I've set the failure threshold for coveralls to allow a 5% decrease (there was none set yet).

src/pom.xml
buehner: ok

buehner:
LGTM, feel free to merge

PR #349, opened 2019-04-02
Update capabilities urls

hwbllmnn:
master version of #348 


hwbllmnn:
The history is just like that because I merged the 1.0.x variant, merging with squash.

PR #352, opened 2019-05-03
Namespace aware handling in default WMS request/response interceptor (1.0.x-master)

dnlkoch:
This fixes the interception of WMS requests and responses in context of the `CUSTOM_ENDPOINT` handling to become more namespace aware:
  * Append `xlink` namespace to element attribute (otherwise clients like QGIS couldn't parse the `GetCapabilities` response).
  * Append the layer's namespace to layername if needed (otherwise we couldn't detect the corresponding GeoServer base URL in subsequent WMS requests as clients like QGIS will make use of the name to generate these requests).
  * Add `@Transactional` annotation to enforce a new JTA transaction if called outside of a transaction context.

Please review @terrestris/devs.


hwbllmnn:
LGTM, please merge

PR #357, opened 2019-05-17
Exclude problematic logging facades

buehner:
As we are not using log4j in v1, we should exclude these facades as they lead to actual problems


buehner:
Yes. I think we'll address the whole logging problematic in the near future

PR #360, opened 2019-05-20
GeoServer Importer: Adds method for updating a layer of a given import task and import job

ahennr:
Title says it all.


buehner:
:+1:

PR #363, opened 2019-06-03
Use other port for http server when testing

buehner:
Port 1234 is used by many of our docker setups, which leads to conflicts when building/testing shogun-core while docker is running.
This will simply change the port from 1234 to 11234


marcjansen:
OK to merge. Wouldn't it be even better if would detect a free port during testing and use that?

PR #368, opened 2019-08-07
Use x-geoserver-credentials header for authorization

hwbllmnn:
Replaces #366 

@terrestris/devs 	Please review.


weskamm:
:+1:

dnlkoch:
I really appreciate this solution compared to the introduction of a new interface! :+1: 

But as far as I see, the default/template interceptor implementations (e.g for the [WmsRequestInterceptor](https://github.com/terrestris/shogun-core/blob/master/src/shogun-core-main/src/main/java/de/terrestris/shoguncore/util/interceptor/impl/WmsRequestInterceptor.java#L47)) wouldn't work (except one would send the header from the client or would implement a custom interceptor adding the `x-geoserver-credentials` header) with this change. Could you please adjust the interceptor implementations to set the appropriate header? But this is non-blocking, feel free to create a separate PR for this.

hwbllmnn:
I adjusted the default interceptors. Is this ok with you?

dnlkoch:
Very nice, thanks. Please merge!

PR #369, opened 2019-08-20
Encode as UTF-8 when posting

buehner:
We had an encoding/umlaut problem in a project, which is solved by encoding the payload correctly.

Please review and merge.


dnlkoch:
Very nice!

Have you checked if we use the default encoding (which is `ISO-8859-1` set in `DEFAULT_CONTENT_CHARSET`) elsewhere? If so, just ignore and merge.

PR #370, opened 2019-08-20
Update and cleanup dependencies

buehner:
This will update a lot of dependencies. A few plugin properties were used twice before, so there was also a bit cleanup. The greenmail test needed an adaption as some auth features have been introduced in the new version.

Please review

### DIFF
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -121,20 +121,20 @@
 
         <webapp-archetype-resource-pom>shogun-core-webapp-archetype/src/main/resources/archetype-resources/pom.xml</webapp-archetype-resource-pom>
 
-        <tomcat.version>9.0.22</tomcat.version>
+        <tomcat.version>9.0.24</tomcat.version>
 
         <!-- Core -->
-        <spring.version>5.1.7.RELEASE</spring.version>
-        <spring-security.version>5.1.5.RELEASE</spring-security.version>
+        <spring.version>5.1.9.RELEASE</spring.version>
+        <spring-security.version>5.1.6.RELEASE</spring-security.version>
         <spring-ldap.version>2.3.2.RELEASE</spring-ldap.version>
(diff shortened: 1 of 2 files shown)


src/pom.xml
```diff
@@ -143,21 +143,21 @@
         <!-- Testing -->
         <junit.version>4.12</junit.version>
         <hamcrest.version>1.3</hamcrest.version>
-        <mockito.version>2.23.0</mockito.version>
+        <mockito.version>2.28.2</mockito.version>
         <jsonpath.version>1.2.0</jsonpath.version>
-        <greenmail.version>1.5.6</greenmail.version>
+        <greenmail.version>1.5.10</greenmail.version>
```
hwbllmnn: I somehow remember I tried that before and it lead to problems. We should keep that in mind if we run into mailing trouble down the road.

src/pom.xml
```diff
@@ -773,7 +770,7 @@
             <dependency>
                 <groupId>com.fasterxml.jackson.core</groupId>
                 <artifactId>jackson-databind</artifactId>
-                <version>${jackson.version}.1</version>
+                <version>${jackson.version}.3</version>
```
hwbllmnn: They still didn't release a proper new version? I only added this sub-sub-sub-version because they only released this single artifact for a security fix and not the whole package...

src/pom.xml
buehner: What kind of problems? The test-problem has been fixed within this MR!? In case of auth problems, the test-fix could also help in your project!? As far as i could see in the release notes and MR of greenmail, the main change is that some kind of auth mechanism has been introduced there

src/pom.xml
buehner: Yes... They did not yet release a proper new version ;-)

PR #381, opened 2020-06-19
Introduces layerappearance clickable property

jansule:
This introduces the `clickable` property for the layerappearance model. The property can be used to set clickable and hoverable configs separately.

This change comes in handy when working with the newly introduced [BasiGX `hoverClick` plugin](https://github.com/terrestris/BasiGX/pull/606).


dnlkoch:
Migrations must be placed in the corresponing project implementation (if needed).

PR #382, opened 2020-06-25
Introduces CreatedHelper

jansule:
Adds a new helper class `CreatedHelper` that allows setting the created field on `PersistentObject`s.

This is needed for the future database performance improvements of shogun2webapp.

Please review @dnlkoch

### DIFF
--- a/src/shogun-core-main/src/main/java/de/terrestris/shoguncore/helper/CreatedHelper.java
+++ b/src/shogun-core-main/src/main/java/de/terrestris/shoguncore/helper/CreatedHelper.java
@@ -1,0 +1,44 @@
+package de.terrestris.shoguncore.helper;
+
+import de.terrestris.shoguncore.model.PersistentObject;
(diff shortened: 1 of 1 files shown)


src/shogun-core-main/src/main/java/de/terrestris/shoguncore/helper/CreatedHelper.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.lang.reflect.Field;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+
+import static org.apache.logging.log4j.LogManager.getLogger;
+
+public class CreatedHelper {
+
+    private static final Logger logger = getLogger(IdHelper.class);
```
dnlkoch, line 15:
```suggestion
    private static final Logger logger = getLogger(CreatedHelper.class);
```

PR #389, opened 2020-07-22
Check for public / anonymous read permission on wmts layers

weskamm:
This allow anonymous users to request WMTS layers that have granted permission to the `public` group.

@devs please review

### DIFF
--- a/src/shogun-core-main/src/main/java/de/terrestris/shoguncore/service/GeoServerInterceptorService.java
+++ b/src/shogun-core-main/src/main/java/de/terrestris/shoguncore/service/GeoServerInterceptorService.java
@@ -1,9 +1,13 @@
 package de.terrestris.shoguncore.service;
 
 import de.terrestris.shoguncore.dao.LayerDao;
+import de.terrestris.shoguncore.dao.UserDao;
+import de.terrestris.shoguncore.model.User;
+import de.terrestris.shoguncore.model.UserGroup;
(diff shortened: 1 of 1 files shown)


src/shogun-core-main/src/main/java/de/terrestris/shoguncore/service/GeoServerInterceptorService.java
```diff
@@ -348,16 +366,38 @@ public Response interceptWmtsRequest(HttpServletRequest request, String serviceI
...
+                if (layer.getSource() instanceof WmtsLayerDataSource) {
+                    WmtsLayerDataSource source = (WmtsLayerDataSource) layer.getSource();
+                    if (source.getId().equals(id)) {
+                        Map<UserGroup, PermissionCollection> userGroupPermissions = layer.getGroupPermissions();
+                        if (userGroupPermissions != null) {
+                            for (UserGroup group : userGroupPermissions.keySet()) {
+                                if (group.getName().equalsIgnoreCase("public") &&
+                                    group.getOwner() == null) {
+                                    dataSource = source;
+                                    break;
```
dnlkoch, line 385: This will break out the inner loop only. Please make either use of labels or move this code to a dedicated function that will return the result value (<- preferred).

weskamm:
closing in favor of https://github.com/terrestris/shogun-core/pull/390

PR #391, opened 2020-08-04
Speedup interceptor by using cached rules

weskamm:
As the title says, will speed up access through the geoserver.action / interceptor by caching the rules that get requested most often.
Speedup is roughly ~300ms decrease per single request.

@devs please review


weskamm:
Thanks for review!

PR #394, opened 2020-08-18
Fix WFS interceptor

hwbllmnn:
Fixes the following issues

* use the custom endpoint/namespace prefix when rewriting URLs in `GetCapabilities` responses
* use the proper WFS 2.0 namespace when removing feature types from `GetCapabilities` responses

### DIFF
--- a/src/shogun-core-main/src/main/java/de/terrestris/shoguncore/util/interceptor/secure/WfsResponseInterceptor.java
+++ b/src/shogun-core-main/src/main/java/de/terrestris/shoguncore/util/interceptor/secure/WfsResponseInterceptor.java
@@ -157,16 +157,15 @@ public class WfsResponseInterceptor implements WfsResponseInterceptorInterface {
             addNamespace("wfs", "http://www.opengis.net/wfs"));
         NodeList list = (NodeList) xpath.compile("//wfs:FeatureType/wfs:Name").evaluate(doc.getDocumentElement(), NODESET);
         List<Element> toRemove = new ArrayList<>();
-        for (int i = 0; i < list.getLength(); ++i) {
-            Element name = (Element) list.item(i);
-            String str = name.getTextContent();
-            if (!layerNames.contains(str)) {
-                toRemove.add((Element) name.getParentNode());
-            }
-        }
+        determineFeatureTypeNodesToRemove(layerNames, list, toRemove);
         xpath.setNamespaceContext(CommonNamespaces.getNamespaceContext().
             addNamespace("wfs", "http://www.opengis.net/wfs/2.0"));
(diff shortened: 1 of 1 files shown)


src/shogun-core-main/src/main/java/de/terrestris/shoguncore/util/interceptor/secure/WfsResponseInterceptor.java
```diff
@@ -157,6 +164,16 @@ private void removeLayers(Document doc) throws XPathExpressionException {
...
             }
         }
+        xpath.setNamespaceContext(CommonNamespaces.getNamespaceContext().
+            addNamespace("wfs", "http://www.opengis.net/wfs/2.0"));
+        list = (NodeList) xpath.compile("//wfs:FeatureType/wfs:Name").evaluate(doc.getDocumentElement(), NODESET);
+        for (int i = 0; i < list.getLength(); ++i) {
+            Element name = (Element) list.item(i);
+            String str = name.getTextContent();
+            if (!layerNames.contains(str)) {
+                toRemove.add((Element) name.getParentNode());
```
weskamm, line 174: you are sure that we should add the parentNode? Line 177 also uses the parentNode, isnt that one level too much?

hwbllmnn: Yes, that's essentially the same code as the existing code, just using a different namespace. L177 only uses the parent node to remove children. For clarity I've factored this out into a helper function in 777a55f7.

weskamm: I was just wondering because the we do 
node->parent1->parent2.remove(parent1)
and thought it should be 
node->parent1.remove(node)

well if you have tested it and it works, feel free to merge.

hwbllmnn: Yeah, the tree looks like this:

```xml
<FeatureTypeList>
  <FeatureType>
    <Name>whatever</Name>
  </FeatureType>
</FeatureTypeList>
````

We fetch the `<Name>` nodes and put the `<FeatureType>` nodes into the `toRemove` list. Then we remove the `<FeatureType>` nodes by calling `removeChild` on the `<FeatureTypeList>` node with the `<FeatureType>` nodes as arguments. I know it looks like traversing too much of the tree, I've often wondered about the missing `remove` method on the actual element you want to remove myself...

weskamm:
Thanks, merging this now!

PR #396, opened 2020-09-10
Disable hibernate second level cache for TreeNodes children

dnlkoch:
This proposes to disable the second level cache for the children of the `TreeNode` as Hibernate won't recognize changes in them and will return the older values always.

First test are very promising. 

Please review @terrestris/devs.


hwbllmnn (approved):

dnlkoch:
Thanks for the review @hwbllmnn!

PR #399, opened 2020-11-27
Fix WFS Transaction in secured interceptor

weskamm:
Adds detection of layer / typenames in xml post bodys, like it happens in WFS transactions.
Also fixed permission detection by handling the ADMIN permission correctly.

### DIFF
--- a/src/shogun-core-main/src/main/java/de/terrestris/shoguncore/util/interceptor/secure/WfsRequestInterceptor.java
+++ b/src/shogun-core-main/src/main/java/de/terrestris/shoguncore/util/interceptor/secure/WfsRequestInterceptor.java
@@ -16,6 +16,7 @@ import de.terrestris.shoguncore.util.interceptor.GeoserverAuthHeaderRequest;
 import de.terrestris.shoguncore.util.interceptor.MutableHttpServletRequest;
 import de.terrestris.shoguncore.util.interceptor.WfsRequestInterceptorInterface;
 
+import org.apache.commons.io.IOUtils;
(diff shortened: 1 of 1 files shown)


src/shogun-core-main/src/main/java/de/terrestris/shoguncore/util/interceptor/secure/WfsRequestInterceptor.java
```diff
@@ -131,6 +134,20 @@ public MutableHttpServletRequest interceptTransaction(MutableHttpServletRequest
     private boolean isAllowed(
         MutableHttpServletRequest request, String paramName, String method) {
         String typeNameParam = request.getParameterIgnoreCase(paramName);
+        if (typeNameParam == null) {
+            // try to get typename from POST body, if any
+            try {
+                String body = IOUtils.toString(request.getInputStream(), StandardCharsets.UTF_8.name());
+                if (!StringUtils.isEmpty(body) && StringUtils.containsIgnoreCase(body, "wfs:transaction")) {
```
marcjansen, line 141: The `wfs` xml namespace can be changed to nearly any string AFAICT. This should probably be enhanced, at least a TODO or follow up issue should be doable, correct?

weskamm: Issue created: https://github.com/terrestris/shogun-core/issues/400

weskamm:
Thanks for review, will merge now

PR #412, opened 2021-09-21
Update dependencies

hwbllmnn:
Updates a lot of plugins and maven dependencies.

### DIFF
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
 
     <modelVersion>4.0.0</modelVersion>
     <groupId>de.terrestris</groupId>
@@ -119,17 +120,18 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
-        <webapp-archetype-resource-pom>shogun-core-webapp-archetype/src/main/resources/archetype-resources/pom.xml</webapp-archetype-resource-pom>
+        <webapp-archetype-resource-pom>shogun-core-webapp-archetype/src/main/resources/archetype-resources/pom.xml
+        </webapp-archetype-resource-pom>
 
         <tomcat.version>9.0.30</tomcat.version>
 
         <!-- Core -->
-        <spring.version>5.2.3.RELEASE</spring.version>
-        <spring-security.version>5.2.2.RELEASE</spring-security.version>
+        <spring.version>5.3.10</spring.version>
+        <spring-security.version>5.5.2</spring-security.version>
         <spring-ldap.version>2.3.2.RELEASE</spring-ldap.version>
-        <log4j.version>2.13.3</log4j.version>
+        <log4j.version>2.14.1</log4j.version>
         <slf4j.version>1.7.28</slf4j.version>
-        <jackson.version>2.10.2</jackson.version>
+        <jackson.version>2.13.0-rc2</jackson.version>
         <opencsv.version>4.6</opencsv.version>
         <ehcache.version>3.8.1</ehcache.version>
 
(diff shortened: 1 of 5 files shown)


src/pom.xml
```diff
@@ -119,17 +120,18 @@
...
-        <spring.version>5.2.3.RELEASE</spring.version>
-        <spring-security.version>5.2.2.RELEASE</spring-security.version>
+        <spring.version>5.3.10</spring.version>
+        <spring-security.version>5.5.2</spring-security.version>
         <spring-ldap.version>2.3.2.RELEASE</spring-ldap.version>
-        <log4j.version>2.13.3</log4j.version>
+        <log4j.version>2.14.1</log4j.version>
         <slf4j.version>1.7.28</slf4j.version>
-        <jackson.version>2.10.2</jackson.version>
+        <jackson.version>2.13.0-rc2</jackson.version>
```
dnlkoch, line 134: Is the rc version required? If not I would prefer to stick to version 2.12.

dnlkoch (approved):

src/pom.xml
hwbllmnn: No, not required. But I also upgraded to this rc version in a couple of other projects and also had no problems there. Should I still downgrade?

dnlkoch: If you didn't encounter any problems, it's absolutely fine for me.

src/shogun-core-main/src/test/java/de/terrestris/shoguncore/web/ImageFileControllerTest.java
```diff
@@ -50,12 +48,9 @@ public void getThumbnailReturnsErrorMapIfIdIsNull() throws Exception {
...
         MvcResult result = mockMvc.perform(MockMvcRequestBuilders.get(GET_THUMBNAIL_ENDPOINT)
             .param("id", ""))
-            .andExpect(status().isOk()).andReturn();
+            .andExpect(status().is4xxClientError()).andReturn();
 
-        assertEquals(result.getResponse().getStatus(), 200);
-        ObjectMapper objectMapper = new ObjectMapper();
-        assertEquals("Returned body matched mocked one.",
-            objectMapper.writeValueAsString(ResultSet.error("Could not get the image thumbnail: Could not find the image with id null")), result.getResponse().getContentAsString());
+        assertEquals(result.getResponse().getStatus(), 400);
```
marcjansen, line 53: This looks a bit as if the API has changed after the upgrades. Is this correct? Would this require a major release?

src/shogun-core-main/src/test/java/de/terrestris/shoguncore/web/MapControllerTest.java
```diff
@@ -61,14 +61,9 @@ public void setLayersForMapTest_erroneous() throws Exception {
         MvcResult result = mockMvc.perform(MockMvcRequestBuilders.post(SETLAYERS_ENDPOINT)
                 .param("mapModuleId", "")
                 .param("layerIds", ""))
-                .andExpect(status().isOk()).andReturn();
-
-        assertEquals(result.getResponse().getStatus(), 200);
+                .andExpect(status().is4xxClientError()).andReturn();
```
marcjansen, line 64: This looks a bit as if the API has changed after the upgrades. Is this correct? Would this require a major release?

marcjansen (commented):
Just one questiin whether we have changed the public API.

src/shogun-core-main/src/test/java/de/terrestris/shoguncore/web/ImageFileControllerTest.java
hwbllmnn: See below, it's the same issue

src/shogun-core-main/src/test/java/de/terrestris/shoguncore/web/MapControllerTest.java
hwbllmnn: It depends on how you look at it. Due to the Spring upgrade, Spring reacts differently if you send it an empty string to a place where an actual mandatory parameter is expected. Previously you'd get a `200` with whatever response the controller produces in this case, now you get a `400` since Spring detects this before giving up control to the controller.

The new behaviour is actually more consistent how Spring works in general, e.g. if you send it a string where a number is expected you'll also get a `400` with the old version.

Now to discuss whether this changes the API, you can surely argue that it does. You could also argue that this actually changes a bug in the API. I prefer to think of the old behaviour as a bug, we're not using Spring to annotate the controller with the required/optional parameters, the allowed content-type, the required HTTP method etc. just to check if all these requirements are met ourselves.

I'll leave it to the next person to make a release to think more about this.

marcjansen: Thanks for the clarification. I'd still think it is a major one. Let's dicuss this then.

How do we ensure we do not forget about this discussion? Adding a label to this issue?

hwbllmnn: I've opened a [discussion](https://github.com/terrestris/shogun-core/discussions/413)

